Issue #19, opened 2016-05-02
Security Descriptors should be preserved in binary form instead of using SDDL


swernli:
Since we recently found some issues with translation of SDDL to binary Security Descriptor when the SDDL includes any domain info, we should skip translating to and from the SDDL and encode the binary directly into the tar streams instead.

PR #20, opened 2016-05-11
Remove use of SDDL in tar headers

swernli:
Switching to using straight binary instead of SDDL in tar headers to avoid failures in converting domain SIDs into/out of SDDL.

Fixes https://github.com/Microsoft/go-winio/issues/19

### DIFF
--- a/wim/wim.go
+++ b/wim/wim.go
@@ -553,6 +553,7 @@ func (r *Reader) readSecurityDescriptors(rsrc io.Reader) (sds [][]byte, n int64,
 		return
 	}
 
+	n = secsize
 	return
 }
 


backuptar/tar.go
```diff
@@ -133,11 +133,7 @@ func WriteTarFileFromBackupStream(t *tar.Writer, r io.Reader, name string, size
 			if err != nil {
 				return err
 			}
-			sddl, err := winio.SecurityDescriptorToSddl(sd)
-			if err != nil {
-				return err
-			}
-			hdr.Winheaders[hdrSecurityDescriptor] = sddl
+			hdr.Winheaders[hdrSecurityDescriptor] = string(sd[:])
```
jstarks: You should base64 encode it so that we're not encoding binary in the string

backuptar/tar.go
```diff
@@ -263,11 +259,8 @@ func FileInfoFromHeader(hdr *tar.Header) (name string, size int64, fileInfo *win
 // tar file that was not processed, or io.EOF is there are no more.
 func WriteBackupStreamFromTarFile(w io.Writer, t *tar.Reader, hdr *tar.Header) (*tar.Header, error) {
 	bw := winio.NewBackupStreamWriter(w)
-	if sddl, ok := hdr.Winheaders[hdrSecurityDescriptor]; ok {
-		sd, err := winio.SddlToSecurityDescriptor(sddl)
-		if err != nil {
-			return nil, err
-		}
+	if sdstr, ok := hdr.Winheaders[hdrSecurityDescriptor]; ok {
```
jstarks: On read, we should support the old way for now as well so that we have compatibility with existing tar files. You'll also want to encode this with something other than "sd" as the header name. (I wish I had chosen "sddl" as the original name).

PR #21, opened 2016-05-11
Remove use of SDDL in tar headers

swernli:
Switching to using straight binary instead of SDDL in tar headers to avoid failures in converting domain SIDs into/out of SDDL.

Fixes https://github.com/Microsoft/go-winio/issues/19

### DIFF
--- a/backuptar/tar.go
+++ b/backuptar/tar.go
@@ -1,6 +1,7 @@
 package backuptar
 
 import (
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


swernli:
Whoops... accidentally pushed a version of the branch with all changes reverted, which seems to have auto-closed the previous PR (https://github.com/Microsoft/go-winio/pull/20).  Rather than try to resurrect, I just opened a new one here.

backuptar/tar.go
```diff
@@ -263,11 +261,20 @@ func FileInfoFromHeader(hdr *tar.Header) (name string, size int64, fileInfo *win
 // tar file that was not processed, or io.EOF is there are no more.
 func WriteBackupStreamFromTarFile(w io.Writer, t *tar.Reader, hdr *tar.Header) (*tar.Header, error) {
 	bw := winio.NewBackupStreamWriter(w)
+	var sd []byte
 	if sddl, ok := hdr.Winheaders[hdrSecurityDescriptor]; ok {
```
jstarks: put a comment in

swernli: Adding...

swernli:
Tested and confirmed correct behavior.

PR #22, opened 2016-05-13
Add function for setting privileges globally

jstarks:
Also optimize privilege setting with the same privilege name multiple
times.


jstarks:
This is a higher priority than the WIM changes.

jterry75:
LGTM

Issue #30, opened 2016-08-25
Add build constraints for non-Windows platforms


rheinwein:
`winio` should not try to compile on non-Windows platforms.

This package is a dependency for many other packages and programs that are intended to build on non-Windows systems.  Because this package does not include any [build constraints](https://golang.org/pkg/go/build/) delcaring which OS it is intended to build on, non-Windows users will receive `syscall` errors, since the `syscall` packages are OS specific.

There are workarounds, but it would be highly preferred for this package to provide build constraints.

https://github.com/docker/swarmkit/issues/1067
https://forums.docker.com/t/problem-while-getting-swarmkit-source/15168

PR #33, opened 2016-10-06
Include build tags to allow go get on non-Windows platforms.

cezarsa:
This PR adds build tags to only try compiling Windows specific code when running on Windows, fixing #30.

As described in #30, this package is vendored in many common libraries like docker and swarmkit and currently it's not possible to run commands like `go get github.com/docker/swarmkit/...` because this library won't compile.

A `noop.go` file was added to the `backuptar` package because at least one file in a package must be compilable on the current platform. Also, the generated `zsyscall.go` was renamed to `zsyscall_windows.go` to avoid having to edit the output from `mksyscall`.


lmars:
Anything preventing this from being merged?

I could really do with this as I am vendoring `github.com/docker/go-connections/sockets` which depends on this library, but I am building on Linux and would like to use `go test ./...` (which currently fails with `undefined: syscall.Overlapped`).

frosenberg:
We see a similar issue with `vendor/github.com/Microsoft/go-winio/file.go:45: undefined: syscall.Overlapped`

darrenstahlmsft:
LGTM

PR #35, opened 2016-11-02
Add DisableProcessPrivileges

jhowardmsft:

@jstarks Adds capability to DisableProcessPrivileges for a process token. Is 0 for disable as found in this sample: https://msdn.microsoft.com/en-us/library/windows/desktop/aa446619(v=vs.85).aspx


jstarks:
LGTM

PR #37, opened 2017-01-09
Replace unnecessary mod17 function with builtin mod

vanita5:
replace pointless(?) modulo function

### DIFF
--- a/wim/lzx/lzx.go
+++ b/wim/lzx/lzx.go
@@ -256,14 +256,6 @@ func (f *decompressor) getCode(h *huffman) uint16 {
 	return 0
 }
 
-// mod17 computes the value mod 17.
-func mod17(b byte) byte {
-	for b >= 17 {
-		b -= 17
-	}
-	return b
-}
-
 // readTree updates the huffman tree path lengths in lens by
 // reading and decoding lengths from the byte stream. lens
 // should be prepopulated with the previous block's tree's path
@@ -288,7 +280,7 @@ func (f *decompressor) readTree(lens []byte) error {
 		}
 		switch {
 		case c <= 16: // length is delta from previous length
-			lens[i] = mod17(lens[i] + 17 - c)
+			lens[i] = (lens[i] + 17 - c) % 17
 			i++
 		case c == 17: // next n + 4 lengths are zero
 			zeroes := int(f.getBits(4)) + 4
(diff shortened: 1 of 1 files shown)


shanemhansen:
This PR should require a benchmark. @vanita5 so politely asks "why?". There may be several possible reasons and frankly none of them deserve snark:

- A function that often only executes a single cmp and jmp can take fewer clock cycles and the instruction used to generate a remainder. If so, aren't you glad we get the faster version?
- This code was ported from C++ code and C++ version ran faster this way. This might not be optimal for the go version. If so, aren't you glad we at least have a pure-go version to iterate on?
- A go novice didn't realize the % operand was available. If so, I welcome this author to the go community!

Source: for the purposes of a friendly Go benchmark competition I have replaced replaced a % operator with a -= for performance. Compression code in tight loops can be tricky.

vanita5:
@shanemhansen This PR is not intended as snark. I'm sorry if you got it that way.

A function to calculate the modulo of a fixed value seems redundant at first sight and I assume it is.

> A function that often only executes a single cmp and jmp can take fewer clock cycles and the instruction used to generate a remainder. If so, aren't you glad we get the faster version?

> This code was ported from C++ code and C++ version ran faster this way. This might not be optimal for the go version.

Well, first of all, talking about C++, I doubt, that this is completely true. If the mod17 function is really faster than the built in % depends mostly on your compiler and platform. And even if it was in some cases, we're talking about micro optimizations here that might not even relevant.
Also, it's really hard to benchmark such tiny operations, so how are you sure this approach actually **is** faster?

> A go novice didn't realize the % operand was available.

That's quite possible. But most modern languages come with a built in % operator. Now we'd have to know if the C++ code it was ported from had this function or not. 

> If so, I welcome this author to the go community!

Yes, you're totally right. Just to make it clear, I don't want to mock on the person who wrote this code. Primarily I'm just interested and I might be totally wrong. Maybe it is this way for a reason.
I'm not part of the Go community, but in this case, I'll welcome the person. Probably both of us can learn something here.

jstarks:
See #38; short answer is that yes, this was hoped to be a small optimization, but either it wasn't measured before or the Go compiler has improved since this was written by adding a strength reduction optimization. Either way, we should eliminate this unnecessary function. Thanks for the PR!

jstarks:
(@vanita5 I can't merge this with the current commit description, though :smile:)

vanita5:
@jstarks Thank you for looking into this. I tried to write a benchmark yesterday and came to the same conclusion.
I'll change the commit description :laughing:

mirabilos:
From what I’ve been seen everywhere, nowadays (~last decade or more), a conditional jump is more expensive than a lot of things, and (for even longer) an integer modulus operation takes one cycle.

Now, if you’re targetting SPARCv8 CPUs, that’s not true, but (other than the one on my coffee table) where would you find them in use? ☺

PR #43, opened 2017-03-17
Flush outgoing content before sending EOF message

### LINKED ISSUE #42: EOF message not received with message mode named pipes
In some circumstances, the 0-length message sent when CloseWrite is called on a message-mode named pipe is not received separately from the previous message. It seems to happen on CloseWrite is called while the previous message has not yet been read by the remote party.

This issue has an impact on Docker CLI and Docker for Windows, when running scripts like:
` echo "hello world" | docker run ... `

simonferquel:
Fixes #42 

This adds a `Flush()` method on `win32File`, and calls it on `CloseWrite()` just before sending the 0-length EOF message.

This makes sure that the 0-length message is not collapsed into the previous message, and that functions like io.Copy() do correctly receive an io.EOF error and completes instead of hanging.

The first commit introduces a test that hangs, the second fixes the issue (and makes the test pass)

This was identified, qualified and fixed by @dgageot and me

### DIFF
--- a/file.go
+++ b/file.go
@@ -219,3 +219,7 @@ func (f *win32File) SetWriteDeadline(t time.Time) error {
 	f.writeDeadline = t
 	return nil
 }
+
+func (f *win32File) Flush() error {
+	return syscall.FlushFileBuffers(f.handle)
+}
--- a/pipe.go
+++ b/pipe.go
@@ -87,7 +87,11 @@ func (f *win32MessageBytePipe) CloseWrite() error {
 	if f.writeClosed {
 		return errPipeWriteClosed
 	}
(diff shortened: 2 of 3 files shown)


thaJeztah:
ping @jstarks PTAL 👍

dgageot:
@jstarks @thaJeztah This is really important for Docker for Windows. Please TAL

vdemeester:
@jstarks @jhowardmsft @darrenstahlmsft we need your review on this 👼 Right now we use a patch version of this to fix some issues with docker4windows but it would be way better if it was fixed upstream 👼 🙏

Issue #44, opened 2017-03-30
Nil pointer dereference in prepareIo


m-kostrzewa:
Sometimes, we get nil pointer derefence in this line in file.go:

```
func (f *win32File) prepareIo() (*ioOperation, error) {
	f.wg.Add(1)								// <---------- HERE
	if f.closing {
		return nil, ErrFileClosed
	}
	c := &ioOperation{}
	c.ch = make(chan ioResult)
	return c, nil
}
```

because ```f``` is nil. This is caused by a race condition in listenerRoutine(), when someone calls pipe.Close() before connectPipe() has a chance to start:

```
p, err := l.makeServerPipe()
if err == nil {
	// Wait for the client to connect.
	ch := make(chan error)
	go func() {
		ch <- connectPipe(p)					// <------ 2. we use niled pointer
	}()
	select {
	case err = <-ch:
		if err != nil {
			p.Close()
			p = nil
		}
	case <-l.closeCh:
		// Abort the connect request by closing the handle.
		p.Close()
		p = nil								// <----- 1. we nil the file pointer before connectPipe()
		err = <-ch
		if err == nil || err == ErrFileClosed {
			err = ErrPipeListenerClosed
		}
		closed = true
	}
```

This seems related to https://github.com/Microsoft/go-winio/pull/31

PR #45, opened 2017-03-30
Fix nil pointer dereference on pipe close

m-kostrzewa:
Fixes https://github.com/Microsoft/go-winio/issues/44

We don't nil the file pointer until after the connectPipe() function returns.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -275,8 +275,8 @@ func (l *win32PipeListener) listenerRoutine() {
 				case <-l.closeCh:
 					// Abort the connect request by closing the handle.
 					p.Close()
-					p = nil
 					err = <-ch
+					p = nil
 					if err == nil || err == ErrFileClosed {
 						err = ErrPipeListenerClosed
 					}


dgageot:
Hi @m-kostrzewa, do you think it's possible to write a unit test that demonstrate the issue?

m-kostrzewa:
Hi @dgageot 

I don't see a way to do this without introducing mocks. I could theoretically write a test that spawns like 1000 goroutines and detect the race that way, but it would be flaky and not really a unit test anymore.

m-kostrzewa:
@dgageot I asked some of my friends and they're not sure how to demonstrate this issue with unit tests either... it's a race condition, after all.

For what it's worth, we merged this patch to our repos 1.5 months ago and the issue never appeared again, and it would happen every other test run before that.

jstarks:
Thanks for the contribution, sorry for the delay. We decided to take PR #63, which should address the same issue.

PR #48, opened 2017-05-01
Added runtime.KeepAlive for the new stricter 1.8 GC

### LINKED ISSUE #41: fatal error: unexpected signal during runtime execution
Hi,

My go binary listens on a named pipe. All works fine normally, but I sometimes get this panic if the named pipe receives a connection:

```
fatal error: unexpected signal during runtime execution
[signal 0xc0000005 code=0x0 addr=0xffffffffffffffff pc=0x40f115]

goroutine 18 [running]:
runtime.throw(0xa53f6f, 0x2a)
	C:/Go/src/runtime/panic.go:596 +0x9c fp=0xc042151e08 sp=0xc042151de8
runtime.sigpanic()
	C:/Go/src/runtime/signal_windows.go:155 +0x18b fp=0xc042151e38 sp=0xc042151e08
runtime.lock(0x5105fcc3b06b83db)
	C:/Go/src/runtime/lock_sema.go:43 +0x65 fp=0xc042151e70 sp=0xc042151e38
runtime.chansend(0x970300, 0x5105fcc3b06b8383, 0xc042151fb0, 0xc042151f01, 0x5d013a, 0xc042151f9c)
	C:/Go/src/runtime/chan.go:176 +0xba fp=0xc042151f20 sp=0xc042151e70
runtime.chansend1(0x970300, 0x5105fcc3b06b8383, 0xc042151fb0)
	C:/Go/src/runtime/chan.go:113 +0x4d fp=0xc042151f60 sp=0xc042151f20
vendor/github.com/Microsoft/go-winio.ioCompletionProcessor(0x1f8)
	src/vendor/github.com/Microsoft/go-winio/file.go:133 +0xea fp=0xc042151fd8 sp=0xc042151f60
runtime.goexit()
	C:/Go/src/runtime/asm_amd64.s:2197 +0x1 fp=0xc042151fe0 sp=0xc042151fd8
created by vendor/github.com/Microsoft/go-winio.initIo
	src/vendor/github.com/Microsoft/go-winio/file.go:57 +0x87
```

Other google results for this error were submitted directly to golang/go, so i don't know if this is an issue with Golang's runtime or an issue with this library.

It seems to happen randomly, based on synchronisation state of other parts of the program.

`go version go1.8 windows/amd64`

**EDIT:** Simplified report using the latest information.

darrenstahlmsft:
fixes #41 

As far as I can tell, this should contain all the runtime.KeepAlive calls needed to conform to the new argument liveness guarantees in Go1.8.x 

It turns out that the main issue was due to passing a Go pointer via native calls to the ioCompletionProcessor, causing the new garbage collection rules to cleanup the struct containing the channel even though it did not have a finalizer.

/cc @jhowardmsft @jstarks PTAL

Thanks @mappu for the initial investigation and repro test.

This also updates zsyscall_windows.go with the new 1.8 go generated code, though it is not related to the KeepAlive, it should be done with the update to 1.8.

### DIFF
--- a/file.go
+++ b/file.go
@@ -79,7 +79,6 @@ func makeWin32File(h syscall.Handle) (*win32File, error) {
 	if err != nil {
 		return nil, err
 	}
-	runtime.SetFinalizer(f, (*win32File).closeHandle)
 	return f, nil
 }
 
@@ -103,7 +102,6 @@ func (f *win32File) closeHandle() {
 // Close closes a win32File.
 func (f *win32File) Close() error {
 	f.closeHandle()
-	runtime.SetFinalizer(f, nil)
 	return nil
 }
 
@@ -166,6 +164,12 @@ func (f *win32File) asyncIo(c *ioOperation, deadline time.Time, bytes uint32, er
 		if wait {
 			r = <-c.ch
 		}
+
+		// runtime.KeepAlive is needed, as c is passed via native
+		// code to ioCompletionProcessor, c must remain alive
+		// until the channel read is complete.
+		runtime.KeepAlive(c)
+
 		err = r.err
 		if err == syscall.ERROR_OPERATION_ABORTED {
 			if f.closing {
@@ -188,6 +192,7 @@ func (f *win32File) Read(b []byte) (int, error) {
 	var bytes uint32
 	err = syscall.ReadFile(f.handle, b, &bytes, &c.o)
 	n, err := f.asyncIo(c, f.readDeadline, bytes, err)
+	runtime.KeepAlive(b)
 
 	// Handle EOF conditions.
 	if err == nil && n == 0 && len(b) != 0 {
@@ -207,7 +212,9 @@ func (f *win32File) Write(b []byte) (int, error) {
 	}
 	var bytes uint32
 	err = syscall.WriteFile(f.handle, b, &bytes, &c.o)
-	return f.asyncIo(c, f.writeDeadline, bytes, err)
+	n, err := f.asyncIo(c, f.writeDeadline, bytes, err)
+	runtime.KeepAlive(b)
+	return n, err
 }
 
 func (f *win32File) SetReadDeadline(t time.Time) error {
--- a/fileinfo.go
+++ b/fileinfo.go
@@ -4,6 +4,7 @@ package winio
 
 import (
 	"os"
+	"runtime"
 	"syscall"
 	"unsafe"
 )
@@ -28,6 +29,7 @@ func GetFileBasicInfo(f *os.File) (*FileBasicInfo, error) {
 	if err := getFileInformationByHandleEx(syscall.Handle(f.Fd()), fileBasicInfo, (*byte)(unsafe.Pointer(bi)), uint32(unsafe.Sizeof(*bi))); err != nil {
 		return nil, &os.PathError{Op: "GetFileInformationByHandleEx", Path: f.Name(), Err: err}
 	}
+	runtime.KeepAlive(f)
 	return bi, nil
 }
 
@@ -36,6 +38,7 @@ func SetFileBasicInfo(f *os.File, bi *FileBasicInfo) error {
 	if err := setFileInformationByHandle(syscall.Handle(f.Fd()), fileBasicInfo, (*byte)(unsafe.Pointer(bi)), uint32(unsafe.Sizeof(*bi))); err != nil {
 		return &os.PathError{Op: "SetFileInformationByHandle", Path: f.Name(), Err: err}
 	}
+	runtime.KeepAlive(f)
 	return nil
 }
 
@@ -52,5 +55,6 @@ func GetFileID(f *os.File) (*FileIDInfo, error) {
 	if err := getFileInformationByHandleEx(syscall.Handle(f.Fd()), fileIDInfo, (*byte)(unsafe.Pointer(fileID)), uint32(unsafe.Sizeof(*fileID))); err != nil {
 		return nil, &os.PathError{Op: "GetFileInformationByHandleEx", Path: f.Name(), Err: err}
 	}
+	runtime.KeepAlive(f)
 	return fileID, nil
 }
--- a/pipe.go
+++ b/pipe.go
@@ -2,6 +2,9 @@
 
 package winio
 
+// #include <stdlib.h>
+import "C"
+
 import (
 	"errors"
 	"io"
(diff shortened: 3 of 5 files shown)


file.go
```diff
@@ -166,6 +166,13 @@ func (f *win32File) asyncIo(c *ioOperation, deadline time.Time, bytes uint32, er
 		if wait {
 			r = <-c.ch
 		}
+
+		// runtime.KeepAlive is needed c and bytes are passed via native
+		// code to ioCompletionProcessor, c and bytes must both remain alive
+		// until the channel read is complete.
+		runtime.KeepAlive(c)
+		runtime.KeepAlive(bytes)
```
jstarks: >bytes [](start = 20, length = 5)

I don't think we need to do anything with bytes here (it's not even a pointer).

file.go
```diff
@@ -221,5 +228,7 @@ func (f *win32File) SetWriteDeadline(t time.Time) error {
 }
 
 func (f *win32File) Flush() error {
-	return syscall.FlushFileBuffers(f.handle)
+	err := syscall.FlushFileBuffers(f.handle)
+	runtime.KeepAlive(f)
```
jstarks: >runtime.KeepAlive(f) [](start = 1, length = 20)

should we just remove the finalizer on win32File?

fileinfo.go
```diff
@@ -36,6 +38,8 @@ func SetFileBasicInfo(f *os.File, bi *FileBasicInfo) error {
 	if err := setFileInformationByHandle(syscall.Handle(f.Fd()), fileBasicInfo, (*byte)(unsafe.Pointer(bi)), uint32(unsafe.Sizeof(*bi))); err != nil {
 		return &os.PathError{Op: "SetFileInformationByHandle", Path: f.Name(), Err: err}
 	}
+	runtime.KeepAlive(f)
+	runtime.KeepAlive(bi)
```
jstarks: >runtime.KeepAlive(bi) [](start = 1, length = 21)

I don't think this is necessary; the cgo rules should keep bi alive across the call to setFileInformationByHandle.

pipe.go
```diff
@@ -231,15 +232,16 @@ func makeServerPipeHandle(path string, securityDescriptor []byte, c *PipeConfig,
...
+	sa.Length = uint32(unsafe.Sizeof(*sa))
 	if securityDescriptor != nil {
 		sa.SecurityDescriptor = &securityDescriptor[0]
 	}
-	h, err := createNamedPipe(path, flags, mode, cPIPE_UNLIMITED_INSTANCES, uint32(c.OutputBufferSize), uint32(c.InputBufferSize), 0, &sa)
+	h, err := createNamedPipe(path, flags, mode, cPIPE_UNLIMITED_INSTANCES, uint32(c.OutputBufferSize), uint32(c.InputBufferSize), 0, sa)
 	if err != nil {
 		return 0, &os.PathError{Op: "open", Path: path, Err: err}
 	}
+	runtime.KeepAlive(sa)
```
jstarks: >runtime.KeepAlive(sa) [](start = 1, length = 21)

ditto here

pipe.go
```diff
@@ -231,15 +232,16 @@ func makeServerPipeHandle(path string, securityDescriptor []byte, c *PipeConfig,
 		mode |= cPIPE_TYPE_MESSAGE
 	}
 
-	var sa securityAttributes
-	sa.Length = uint32(unsafe.Sizeof(sa))
+	sa := &securityAttributes{}
+	sa.Length = uint32(unsafe.Sizeof(*sa))
 	if securityDescriptor != nil {
 		sa.SecurityDescriptor = &securityDescriptor[0]
```
jstarks: >&securityDescriptor[0] [](start = 26, length = 22)

This actually violates cgo rules: [https://golang.org/cmd/cgo/](https://golang.org/cmd/cgo/)

Probably we should be duplicating this into a C-allocated buffer.

jstarks:
> 	err = syscall.WriteFile(f.handle, b, &bytes, &c.o)

I'm worried that b could be collected before asyncIo returns... I think we need to keep it alive explicitly. I'm not sure if KeepAlive directly on a slice does anything or if you need KeepAlive on a pointer to the first element of the slice.

(Technically this violates cgo rules, but we don't have an alternative other than copying the data to a C buffer before doing the WriteFile.)

---
Refers to: file.go:216 in bb2dfe9. [](commit_id = bb2dfe93576afa803b2fd531b930ab925e94664b, deletion_comment = False)

jstarks:
> 	err = syscall.WriteFile(f.handle, b, &bytes, &c.o)

And of course we need to do this everywhere else we call asyncIo as well.

---
In reply to: [298452715](https://github.com/Microsoft/go-winio/pull/48#issuecomment-298452715) [](ancestors = 298452715)

---
Refers to: file.go:216 in bb2dfe9. [](commit_id = bb2dfe93576afa803b2fd531b930ab925e94664b, deletion_comment = False)

jstarks:
:clock1:

fileinfo.go
darrenstahlmsft: This one I'm unsure about. After `(*byte)(unsafe.Pointer(bi))` returns the pointer, is `bi` still considered live? How do the new runtime argument liveness rules apply to cgo rules? `bi` is no longer accessible as soon as `unsafe.Pointer` returns, so based on my understanding, a finalizer could run on `bi` before the call to c.

It doesn't matter though, as `bi` does not have a finalizer on it, so I can remove this regardless, but I don't see any documents regarding the new liveness rules when passing pointers to cgo.

jstarks: unsafe.Pointer(x) and (*byte)(x) are casts, not function calls, so they're presumably handled differently. So I don't think that's true.

file.go
darrenstahlmsft: That works for me.

darrenstahlmsft: Also removing BackupFileReader finalizer

fileinfo.go
darrenstahlmsft: Ah, ok. Removed.

darrenstahlmsft:
Tested via manual GC invocations, `b` is fair game for the GC as soon as WriteFile returns. `runtime.KeepAlive(b)` prevents the cleanup of the whole slice and backing array.

Adding the KeepAlives

darrenstahlmsft:
Updated.

file.go
```diff
@@ -221,5 +228,6 @@ func (f *win32File) SetWriteDeadline(t time.Time) error {
 }
 
 func (f *win32File) Flush() error {
-	return syscall.FlushFileBuffers(f.handle)
+	err := syscall.FlushFileBuffers(f.handle)
+	return err
```
jstarks: unnecessary

jstarks:
Looks good after a minor tweak and separating the CBytes thing out as a separate commit.

darrenstahlmsft:
Split into two commits and nit fixed.

PR #50, opened 2017-05-04
Apply read and write deadlines to pending IO

darrenstahlmsft:
In order to upgrade to go1.8 all IO operations including previous pending IO must have the deadline set when SetReadDeadline or SetWriteDeadline is called.

This moves the timeout cancelation responsibility out of the wait routine and into timers in the deadline code.

@jstarks PTAL

### DIFF
--- a/file.go
+++ b/file.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"runtime"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 )
@@ -16,6 +17,12 @@ import (
 //sys getQueuedCompletionStatus(port syscall.Handle, bytes *uint32, key *uintptr, o **ioOperation, timeout uint32) (err error) = GetQueuedCompletionStatus
 //sys setFileCompletionNotificationModes(h syscall.Handle, flags uint8) (err error) = SetFileCompletionNotificationModes
 //sys timeBeginPeriod(period uint32) (n int32) = winmm.timeBeginPeriod
+
+type atomicBool int32
+
+func (b *atomicBool) isSet() bool { return atomic.LoadInt32((*int32)(b)) != 0 }
+func (b *atomicBool) setFalse()   { atomic.StoreInt32((*int32)(b), 0) }
+func (b *atomicBool) setTrue()    { atomic.StoreInt32((*int32)(b), 1) }
 
 const (
 	cFILE_SKIP_COMPLETION_PORT_ON_SUCCESS = 1
@@ -32,6 +39,8 @@ type timeoutError struct{}
 func (e *timeoutError) Error() string   { return "i/o timeout" }
 func (e *timeoutError) Timeout() bool   { return true }
 func (e *timeoutError) Temporary() bool { return true }
+
+type timeoutChan chan struct{}
 
 var ioInitOnce sync.Once
 var ioCompletionPort syscall.Handle
@@ -63,8 +72,16 @@ type win32File struct {
 	handle        syscall.Handle
 	wg            sync.WaitGroup
 	closing       bool
-	readDeadline  time.Time
-	writeDeadline time.Time
+	readDeadline  deadlineHandler
+	writeDeadline deadlineHandler
+}
+
+type deadlineHandler struct {
+	setLock     sync.Mutex
+	channel     timeoutChan
+	channelLock sync.RWMutex
+	timer       *time.Timer
+	timedout    atomicBool
 }
 
 // makeWin32File makes a new win32File from an existing file handle
@@ -79,6 +96,8 @@ func makeWin32File(h syscall.Handle) (*win32File, error) {
 	if err != nil {
 		return nil, err
 	}
+	f.readDeadline.channel = make(timeoutChan)
+	f.writeDeadline.channel = make(timeoutChan)
 	return f, nil
 }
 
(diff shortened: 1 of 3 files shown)


file.go
```diff
@@ -16,6 +17,13 @@ import (
 //sys getQueuedCompletionStatus(port syscall.Handle, bytes *uint32, key *uintptr, o **ioOperation, timeout uint32) (err error) = GetQueuedCompletionStatus
 //sys setFileCompletionNotificationModes(h syscall.Handle, flags uint8) (err error) = SetFileCompletionNotificationModes
 //sys timeBeginPeriod(period uint32) (n int32) = winmm.timeBeginPeriod
+//sys LocalAlloc(uFlags uint32, uBytes uintptr) (mem uintptr) = LocalAlloc
```
jstarks: not sure we should export this... maybe `localAlloc` is better.

darrenstahlmsft: Oops. Alloc is not needed for this change.

jstarks:
We should instead keep the model where `asyncIo` cancels the IO it owns rather than have `SetReadDeadline` do it; this keeps the ownership model for the overlapped structure simple. I suggest having two channels on the file to wait on (for reads and for writes) and signal these from `SetReadDeadline`. I like the idea of using a single timer per deadline instead of one per `asyncIo` call.

darrenstahlmsft:
Updated to move back to that model. I still had to add a lock around the channel though, as the "close channel + create new channel" operation needs to be atomic to prevent a race between the readers and the writer.

jstarks:
> 		runtime.KeepAlive(c)

may as well move this out of the select to the bottom

---
Refers to: file.go:169 in 6acdbc8. [](commit_id = 6acdbc8466ebea825bb83e541a192bf077e08b5f, deletion_comment = False)

file.go
```diff
@@ -174,13 +172,21 @@ func (f *win32File) asyncIo(c *ioOperation, deadline time.Time, bytes uint32, er
...
+	case <-timeout:
+		cancelIoEx(f.handle, &c.o)
+
+		<-c.ch
+		// runtime.KeepAlive is needed, as c is passed via native
+		// code to ioCompletionProcessor, c must remain alive
+		// until the channel read is complete.
+		runtime.KeepAlive(c)
+
+		err = ErrTimeout
```
jstarks: >err = ErrTimeout [](start = 2, length = 16)

you still need to look at r.err here -- it's possible that the IO completed successfully or with a different error even though you called CancelIoEx

file.go
```diff
@@ -210,20 +222,78 @@ func (f *win32File) Write(b []byte) (int, error) {
...
+	}
 	return nil
 }
 
-func (f *win32File) SetWriteDeadline(t time.Time) error {
-	f.writeDeadline = t
+func (f *win32File) SetWriteDeadline(deadline time.Time) error {
+	now := time.Now()
+	duration := deadline.Sub(now)
+
```
jstarks: can move this to after the IsZero check

file.go
```diff
@@ -210,20 +222,78 @@ func (f *win32File) Write(b []byte) (int, error) {
...
+	if f.writeTimer != nil {
+		f.writeTimer.Stop()
+	}
+	f.writeTimer = nil
```
jstarks: >.writeTimer = nil [](start = 2, length = 17)

move into the if

file.go
```diff
@@ -134,37 +149,20 @@ func ioCompletionProcessor(h syscall.Handle) {
 
 // asyncIo processes the return value from ReadFile or WriteFile, blocking until
 // the operation has actually completed.
-func (f *win32File) asyncIo(c *ioOperation, deadline time.Time, bytes uint32, err error) (int, error) {
+func (f *win32File) asyncIo(c *ioOperation, timeout timeoutChan, bytes uint32, err error) (int, error) {
```
jstarks: >timeout timeoutChan [](start = 44, length = 19)

better to make this a bool and get the channel just before the select. this way, you can grab a mutex to get the channel, and you can avoid taking the mutex for IOs that complete synchronously

file.go
```diff
@@ -210,20 +222,78 @@ func (f *win32File) Write(b []byte) (int, error) {
...
+
+	if deadline.IsZero() {
+		return nil
+	}
+
+	timeoutWrites := func() {
+		f.writeTimedOut.setTrue()
+		close(f.writeTimeoutChan)
+		// TODO sync here???
+		f.writeTimeoutChan = make(chan struct{})
```
jstarks: >f.writeTimeoutChan = make(chan struct{}) [](start = 2, length = 40)

I don't think you want to replace the closed channel until someone calls SetDeadline again... You might be able to get away with it if you change asyncIo to re-check writeTimedOut in the mutex (which doesn't exist yet) while grabbing the channel... But maybe it's easier to just leave the channel closed until someone changes the deadline.

jstarks:
I don't see the lock yet. Missing an update?

---
In reply to: [299528825](https://github.com/Microsoft/go-winio/pull/50#issuecomment-299528825) [](ancestors = 299528825)

jstarks:
You should add some tests.

file.go
darrenstahlmsft: asyncIO would need to be made aware of if it's in a read or a write to choose which channel to wait on. Doable, but not sure if it makes things more complicated than needed.

file.go
darrenstahlmsft: This needs to be replaced right away. If a new read or write comes in while there is no channel to wait on it will never be able to cancel a pending IO (A closed channel cannot be reopened, so it can't just wait on the closed placeholder channel)

darrenstahlmsft:
Ah, forgot to push the last change. Updating addressing comments and adding mutex.

file.go
jstarks: I don't understand. If a new read or write comes in, it would wait on the closed channel which should complete immediately. That's what we want, right?

file.go
jstarks: I think we should avoid taking the lock unless we actually go async, so I think this will be necessary.

darrenstahlmsft:
Updated. Still writing tests.

file.go
darrenstahlmsft: Oh. I get it now, yep. Fixed.

darrenstahlmsft:
Moved lock to asyncIO, and combined some duplicated code with a deadlineHandler object.

file.go
```diff
@@ -63,8 +72,15 @@ type win32File struct {
 	handle        syscall.Handle
 	wg            sync.WaitGroup
 	closing       bool
-	readDeadline  time.Time
-	writeDeadline time.Time
+	readDeadline  *deadlineHandler
```
jstarks: >readDeadline  *deadlineHandler [](start = 1, length = 30)

no reason to make this a separate allocation

file.go
```diff
@@ -210,23 +233,61 @@ func (f *win32File) Write(b []byte) (int, error) {
...
+	return f.writeDeadline.set(deadline)
 }
 
 func (f *win32File) Flush() error {
 	return syscall.FlushFileBuffers(f.handle)
 }
+
+func (d *deadlineHandler) set(deadline time.Time) error {
+	select {
+	case <-d.channel:
```
jstarks: seems like there's a race here where the existing timer _just_ expired...

file.go
```diff
@@ -210,23 +233,61 @@ func (f *win32File) Write(b []byte) (int, error) {
...
+	}
+	d.timedout.setFalse()
+
+	if deadline.IsZero() {
+		return nil
+	}
+
+	timeoutIO := func() {
+		d.timedout.setTrue()
+		close(d.channel)
```
jstarks: >d.channel [](start = 8, length = 9)

probably need to hold the lock to grab d.channel

file.go
```diff
@@ -210,23 +233,61 @@ func (f *win32File) Write(b []byte) (int, error) {
...
+	select {
+	case <-d.channel:
+		d.channelLock.Lock()
+		d.channel = make(chan struct{})
+		d.channelLock.Unlock()
+	default:
+	}
+
+	if d.timer != nil {
+		d.timer.Stop()
```
jstarks: >Stop [](start = 10, length = 4)

I wonder if you can use the return value from Stop to clean up some of the race conditions here

file.go
jstarks: Maybe you can avoid this if you ensure the timer is completely stopped before changing the channel.

file.go
darrenstahlmsft: Yep, that cleans up the races with the timer I think. We should support concurrent calls to SetDeadline though, so I think a deadlineHandler lock is needed

file.go
darrenstahlmsft: Avoided by waiting on channel close and timer stopped before writing to it

darrenstahlmsft:
Fixed timer races

darrenstahlmsft:
Added tests in the first commit, fixed in second.

jstarks:
LGTM

PR #51, opened 2017-05-05
Remove cgo as a dependency added by #49

### LINKED ISSUE #49: Re-evaluate "cgo" dependency?
As of [23ad5a714ce953959ea79723df23a913ff886ddd](https://github.com/Microsoft/go-winio/commit/23ad5a714ce953959ea79723df23a913ff886ddd), the `go-winio` package depends on `cgo`. Some projects, mine included, try to avoid `cgo` dependencies because of the can of worms it opens, including adding complexity to cross-compiling and, in my case, the addition of the `gcc` toolchain to builder systems.

If `cgo` is required, so be it, but I figured I'd at least raise the issue for discussion. Is there any way to achieve a similar end without it? Thanks!

darrenstahlmsft:
fixes #49 

#48 introduced CGO as a dependency when it wasn't really needed (Surprised we didn't already have a CGO dependency tbh). This removes said dependency by delegating the memory management to Windows.

/cc @schmichael @jstarks 

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -2,9 +2,6 @@
 
 package winio
 
-// #include <stdlib.h>
-import "C"
-
 import (
 	"errors"
 	"io"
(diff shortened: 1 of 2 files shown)


jstarks:
LGTM

PR #53, opened 2017-05-19
Move f.wg.Done() to Read and Write to avoid leaking handles

darrenstahlmsft:
This fixes a handle leak introduced in #50 

@jstarks 


darrenstahlmsft:
Updated to include pipe.go and a note on prepareIO that `The caller must call f.wg.Done() when the IO is finished, prior to calling Close().`

PR #54, opened 2017-05-24
RtlCopyMemory is not available on Nanoserver use builtin copy instead

darrenstahlmsft:

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -13,19 +13,12 @@ import (
 )
 
 //sys connectNamedPipe(pipe syscall.Handle, o *syscall.Overlapped) (err error) = ConnectNamedPipe
-//sys createNamedPipe(name string, flags uint32, pipeMode uint32, maxInstances uint32, outSize uint32, inSize uint32, defaultTimeout uint32, sa *securityAttributes) (handle syscall.Handle, err error)  [failretval==syscall.InvalidHandle] = CreateNamedPipeW
-//sys createFile(name string, access uint32, mode uint32, sa *securityAttributes, createmode uint32, attrs uint32, templatefile syscall.Handle) (handle syscall.Handle, err error) [failretval==syscall.InvalidHandle] = CreateFileW
+//sys createNamedPipe(name string, flags uint32, pipeMode uint32, maxInstances uint32, outSize uint32, inSize uint32, defaultTimeout uint32, sa *syscall.SecurityAttributes) (handle syscall.Handle, err error)  [failretval==syscall.InvalidHandle] = CreateNamedPipeW
+//sys createFile(name string, access uint32, mode uint32, sa *syscall.SecurityAttributes, createmode uint32, attrs uint32, templatefile syscall.Handle) (handle syscall.Handle, err error) [failretval==syscall.InvalidHandle] = CreateFileW
 //sys waitNamedPipe(name string, timeout uint32) (err error) = WaitNamedPipeW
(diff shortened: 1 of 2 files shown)


pipe.go
```diff
@@ -18,12 +18,10 @@ import (
 //sys waitNamedPipe(name string, timeout uint32) (err error) = WaitNamedPipeW
 //sys getNamedPipeInfo(pipe syscall.Handle, flags *uint32, outSize *uint32, inSize *uint32, maxInstances *uint32) (err error) = GetNamedPipeInfo
 //sys getNamedPipeHandleState(pipe syscall.Handle, state *uint32, curInstances *uint32, maxCollectionCount *uint32, collectDataTimeout *uint32, userName *uint16, maxUserNameSize uint32) (err error) = GetNamedPipeHandleStateW
-//sys localAlloc(uFlags uint32, length uint32) (ptr uintptr) = LocalAlloc
-//sys copyMemory(dst uintptr, src uintptr, length uint32) = RtlCopyMemory
 
 type securityAttributes struct {
```
jstarks: I wonder why we aren't just using syscall.SecurityAttributes.

pipe.go
```diff
@@ -18,12 +18,10 @@ import (
...
 	Length             uint32
-	SecurityDescriptor uintptr
+	SecurityDescriptor []byte
```
jstarks: This won't work.

pipe.go
```diff
@@ -237,9 +235,8 @@ func makeServerPipeHandle(path string, securityDescriptor []byte, c *PipeConfig,
 	sa.Length = uint32(unsafe.Sizeof(*sa))
 	if securityDescriptor != nil {
 		len := uint32(len(securityDescriptor))
-		sa.SecurityDescriptor = localAlloc(0, len)
```
jstarks: Still need the localAlloc, I think.

pipe.go
darrenstahlmsft: Didn't know that existed. Using it now.

pipe.go
darrenstahlmsft: It works. Likely due to a fluke, but this is removed now

pipe.go
darrenstahlmsft: Wasn't thinking. Added it back.

darrenstahlmsft:
This is closer to the correct fix here. I'm still unsure about the casts on `sa.SecurityDescriptor`, and `go vet` is complaining, but it currently works.

pipe.go
```diff
@@ -233,13 +226,13 @@ func makeServerPipeHandle(path string, securityDescriptor []byte, c *PipeConfig,
...
 
-	sa := &securityAttributes{}
+	sa := &syscall.SecurityAttributes{}
 	sa.Length = uint32(unsafe.Sizeof(*sa))
 	if securityDescriptor != nil {
 		len := uint32(len(securityDescriptor))
 		sa.SecurityDescriptor = localAlloc(0, len)
 		defer localFree(sa.SecurityDescriptor)
-		copyMemory(sa.SecurityDescriptor, uintptr(unsafe.Pointer(&securityDescriptor[0])), len)
+		copy((*[0xffff]byte)(unsafe.Pointer(sa.SecurityDescriptor))[:len], (*[0xffff]byte)(unsafe.Pointer(&securityDescriptor[0]))[:len])
```
jstarks: You only need to cast the destination. The source is already a `[]byte` so it can be passed directly to `copy`.

darrenstahlmsft: Fixed

jstarks:
LGTM

PR #58, opened 2017-07-19
Move vhd to root

jhowardmsft:

I need to (at least temporarily) move the VHD functions into the main package. Need to dig in more and understand once I'm back from vacation, but dockers vndr tool isn't picking up things under vhd\, hence I'm blocked on the dynamic sandbox management PR right now. IOW, a temporary hack.

@jstarks


jhowardmsft:
NVM. figured it.

PR #59, opened 2017-07-19
Bool is not atomic

timou:
The Go race detector doesn't like concurrent access to bool. This change
secures the bool with a mutex.

### DIFF
--- a/file.go
+++ b/file.go
@@ -23,6 +23,13 @@ type atomicBool int32
 func (b *atomicBool) isSet() bool { return atomic.LoadInt32((*int32)(b)) != 0 }
 func (b *atomicBool) setFalse()   { atomic.StoreInt32((*int32)(b), 0) }
 func (b *atomicBool) setTrue()    { atomic.StoreInt32((*int32)(b), 1) }
+func (b *atomicBool) swap(new bool) bool {
+	var newInt int32
+	if new {
+		newInt = 1
+	}
+	return atomic.SwapInt32((*int32)(b), newInt) == 1
+}
 
 const (
 	cFILE_SKIP_COMPLETION_PORT_ON_SUCCESS = 1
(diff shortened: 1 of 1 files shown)


darrenstahlmsft:
We already have an "atomicBool" type in this file used for a different bool with a similar issue. I'd prefer to use that instead of adding a Mutex.

```
type atomicBool int32

func (b *atomicBool) isSet() bool { return atomic.LoadInt32((*int32)(b)) != 0 }
func (b *atomicBool) setFalse()   { atomic.StoreInt32((*int32)(b), 0) }
func (b *atomicBool) setTrue()    { atomic.StoreInt32((*int32)(b), 1) }
```

timou:
That's fine, and I'm happy to change that, but I also think there is a genuine race between inspecting the bool and setting it (lines 112-114 of file.go). I don't think that can be solved by atomicbool.

timou:
OK, I lie. I have extended atomicBool with compareAndSwap, which I think fixes the race and avoids the mutex.

file.go
```diff
@@ -107,9 +117,9 @@ func MakeOpenFile(h syscall.Handle) (io.ReadWriteCloser, error) {
 
 // closeHandle closes the resources associated with a Win32 handle
 func (f *win32File) closeHandle() {
-	if !f.closing {
+	swapped := f.closing.compareAndSwap(false, true)
+	if swapped {
```
darrenstahlmsft: This can be all one line.

darrenstahlmsft:
@timou I was about to comment almost the same fix.

While compareAndSwap works, as per conversation with @jstarks only a swap is needed here (and is faster).

```
func (b *atomicBool) swap(new bool) (old bool) {
	var newInt int32
	if new {
		newInt = 1
	}
	return atomic.SwapInt32((*int32)(b), newInt) == 1
}
```

darrenstahlmsft:
There's still a second (blank) merge commit. I'm good to merge if you can remove that.

timou:
OK, rebased. Thanks for the nice review!

PR #61, opened 2017-07-19
Don't increment the wait count if the file is closing

mlaventure:
This prevent a deadlock if `prepareIo()` is called on a closing file as
`Done()` would never been called afterwards.

Signed-off-by: Kenfe-Mickael Laventure <mickael.laventure@gmail.com>


timou:
+1 for this change. I have a test where I rapidly open and close connections, and it occasionally ends up blocked forever in Close. I've rerun the test dozens of times with this change and it appears to fix the issue.

darrenstahlmsft:
LGTM

Issue #29, opened 2016-07-14
Misuse of sync.WaitGroup


fjl:
With the test below, the following crash occurs instantly. The test is a bit
unwieldy, sorry about that. I am using Go 1.6 on Windows 10.
The documentation of package sync states that

>  If a WaitGroup is reused to wait for several independent sets of events, new Add calls must happen after all previous Wait calls have returned.

The runtime check for this invariant was added in Go 1.5.
I wonder why this issue hasn't come up yet.

``` text
=== RUN   TestCloseRace
panic: sync: WaitGroup is reused before previous Wait has returned

goroutine 369 [running]:
panic(0x572cc0, 0xc08209b460)
    C:/tools/go/src/runtime/panic.go:464 +0x3f4
sync.(*WaitGroup).Wait(0xc0820e3638)
    C:/tools/go/src/sync/waitgroup.go:129 +0x11b
github.com/microsoft/go-winio.(*win32File).closeHandle(0xc0820e3630)
    c:/Users/fjl/src/github.com/microsoft/go-winio/file.go:94 +0x62
github.com/microsoft/go-winio.(*win32File).Close(0xc0820e3630, 0x0, 0x0)
    c:/Users/fjl/src/github.com/microsoft/go-winio/file.go:103 +0x34
github.com/microsoft/go-winio.TestCloseRace.func1.1()
    c:/Users/fjl/src/github.com/microsoft/go-winio/pipe_test.go:293 +0x12d
created by time.goFunc
    C:/tools/go/src/time/sleep.go:129 +0x41
exit status 2
```

The test:

``` go
func TestCloseRace(t *testing.T) {
    l, err := ListenPipe(testPipeName, &PipeConfig{})
    if err != nil {
        t.Fatal(err)
    }

    var (
        maxCloseTimeout   = 100 * time.Millisecond
        numClients        = 10
        numConnsPerClient = 10
        writeDelay        = 30 * time.Millisecond
        writeData         = make([]byte, 100)
        wg                sync.WaitGroup
    )
    server := func() {
        defer wg.Done()
        defer l.Close()
        for {
            conn, err := l.Accept()
            if err != nil {
                t.Log("accept error:", err)
                return
            }
            // Kill the connection after a random delay.
            timeout := time.Duration(rand.Int63n(int64(maxCloseTimeout)))
            time.AfterFunc(timeout, func() {
                t.Logf("killing connection after %v", timeout)
                conn.Close()
            })
[...]

PR #65, opened 2017-08-03
Prevent adding to a sync.WaitGroup that has a waiter

darrenstahlmsft:
fixes https://github.com/Microsoft/go-winio/issues/29

Go does not allow adding to a sync.WaitGroup that currently has a waiter. This added lock prevents the (rare) case that it happens.

This PR also includes a small test bug fix that was found while working on this issue.

There is no test because the repro conditions are very unreliable, and the test I've been using takes 30s to run (10 times that of the entire current test suite) and only catches it in the race detector at that.

@jstarks

### DIFF
--- a/file.go
+++ b/file.go
@@ -78,6 +78,7 @@ func initIo() {
 type win32File struct {
 	handle        syscall.Handle
 	wg            sync.WaitGroup
+	wgLock        sync.RWMutex
 	closing       atomicBool
 	readDeadline  deadlineHandler
 	writeDeadline deadlineHandler
@@ -118,7 +119,9 @@ func (f *win32File) closeHandle() {
 	if !f.closing.swap(true) {
 		// cancel all IO and wait for it to complete
 		cancelIoEx(f.handle, nil)
+		f.wgLock.Lock()
 		f.wg.Wait()
+		f.wgLock.Unlock()
 		// at this point, no new IO can start
 		syscall.Close(f.handle)
 		f.handle = 0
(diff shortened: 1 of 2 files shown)


file.go
```diff
@@ -134,10 +137,13 @@ func (f *win32File) Close() error {
 // prepareIo prepares for a new IO operation.
 // The caller must call f.wg.Done() when the IO is finished, prior to Close() returning.
 func (f *win32File) prepareIo() (*ioOperation, error) {
+	f.wgLock.RLock()
 	if f.closing.isSet() {
```
darrenstahlmsft: Note I chose to keep the atomic bool here as it is accessed in multiple locations that are not affected by this race and think that adding the lock around each access would likely be higher perf impact than an atomic bool and lock in this one case. I didn't test this theory though, so I might be wrong.

file.go
```diff
@@ -118,7 +119,9 @@ func (f *win32File) closeHandle() {
 	if !f.closing.swap(true) {
 		// cancel all IO and wait for it to complete
 		cancelIoEx(f.handle, nil)
+		f.wgLock.Lock()
```
jstarks: >f.wgLock.Lock() [](start = 2, length = 15)

I expected this lock to be around the modification of f.closing, not the wait on the wait group... With this, a new IO on a closing handle will stall until all existing IOs complete.

jstarks:
:clock1:

darrenstahlmsft:
Updated.

PR #70, opened 2018-01-09
pipe: Resolve race between close and connect

jstarks:
This change resolves the issue where Accept() returns ERROR_NO_DATA
because the client closes the connection immediately. It resolves
the race by ignoring that particular connection and waiting on a
new one.


jstarks:
This should fix moby/moby#34207. @jhowardmsft @darrenstahlmsft

jhowardmsft:
LGTM. I'll setup another repro to verify 34207 too.

Issue #69, opened 2018-01-09
Forcing of Windows timer resolution to 1ms no longer necessary, degrades performance under go1.9


justfalter:
It is my belief that changes introduced by #12 and #13 should be reverted. Reasons for doing so:
- go-winio forces the Windows timer to 1ms resolution, which has a significant impact on battery life, as well as causes noticeable CPU utilization for virtual machines. 
- This forcing of the Windows timer to 1ms resolution was done to work around an issue pertaining _only_ to Go 1.6 (#12 and #13), however go-winio no longer builds on Go 1.6. 
- go-winio's forcing of the high-resolution timer defeats improvements to Windows timer handling introduced in Go 1.9.

Background:
- Go 1.6 removed a call that set the Windows timer to 1ms resolution (timeBeginPeriod(1)). https://golang.org/doc/go1.6 :
> On Windows, Go programs in Go 1.5 and earlier forced the global Windows timer resolution to 1ms at startup by calling timeBeginPeriod(1). Go no longer needs this for good scheduler performance, and changing the global timer resolution caused problems on some systems, so the call has been removed.
- This negatively affected performance of go-winio, so #12 and #13 were introduced as a workaround to set the timer resolution to 1ms. 
- Go 1.7 returned to the behavior of Go 1.5, and once again set the timer to 1ms resolution. https://golang.org/doc/go1.7 :
> On Windows, Go programs in Go 1.5 and earlier forced the global Windows timer resolution to 1ms at startup by calling timeBeginPeriod(1). Changing the global timer resolution caused problems on some systems, and testing suggested that the call was not needed for good scheduler performance, so Go 1.6 removed the call. Go 1.7 brings the call back: under some workloads the call is still needed for good scheduler performance.
- Go 1.9 further refined the behavior https://golang.org/doc/go1.9:
> On Windows, Go no longer forces the system timer to run at high resolution when the program is idle. This should reduce the impact of Go programs on battery life.
- By always forcing the timer to 1ms, go-winio effectively defeats the improvements in timer handling made with go1.9. This leads to a significant impact on battery life, as well as VM performance. 

I have found that, under go1.9, commenting out the line that sets the Windows timer to 1ms (https://github.com/Microsoft/go-winio/blob/master/file.go#L157) reduces VM host CPU utilization by ~25%, while having no detectable impact on named-pipe performance.

Note: I would contribute a pull request, myself, but I am currently unable to sign the CLA.

[...]

PR #13, opened 2016-03-15
Fixed error in #12

darrenstahlmsft:
Oops, capitalization error on last PR. This fixes it @jstarks 

Signed-off-by: Darren Stahl <darst@microsoft.com>

PR #12, opened 2016-03-15
Fixes performance regression in GO 1.6

darrenstahlmsft:
Signed-off-by: Darren Stahl <darst@microsoft.com>

PR #71, opened 2018-01-09
Remove workaround to perf issues in go1.6

darrenstahlmsft:
Starting in go1.9, Golang lowers the timer frequency when not required reducing CPU usage. The fix to a perf regression in #12 and #13 was a workaround that prevents Golang from doing so. This PR removes the workaround, since the initial issue was fixed. I see no perf regressions in the test suite.

This does not need to be conditional based on Golang verison, as the issue being worked around was only in go1.6, which is no longer supported by go-winio.

Fixes https://github.com/Microsoft/go-winio/issues/69


justfalter:
👍  Look forward to this getting merged.

darrenstahlmsft:
Merged 😄  Thanks for the report @justfalter

darrenstahlmsft:
I've also tagged a new release `v0.4.7` with the included fix.

PR #72, opened 2018-02-20
Avoid deadlock in Close()

timou:
A call to listenerRoutine while closing can result in a deadlock. The
cause appears to be listenerRoutine not giving up when faced with an
error of type ERROR_OPERATION_ABORTED.

The change adds a case to specifically handle this error type and bubble
up ErrFileClosed to listenerRoutine, causing it to terminate gracefully.


gdamore:
This change looks straight-forward, and good.  Hopefully it gets integrated soon.

timou:
This change has been sitting around a while. Any feedback?

StefanoBalzarottiNozomi:
Any news? @helsaawy

PR #74, opened 2018-02-28
Add option to accept remote named pipe connections

nnam:
Adds an optional field to the PipeConfig struct that toggles whether or not a named pipe can accept remote connections.


h4ng3r:
Hi we are using go-winio on a project and had to fork it because it was not accepting remote connections. Could you please refactor and approve this PR? Thanks

PR #78, opened 2018-05-25
Allow for message pipes in message readmode for Windows named pipes to work.

mnltejaswini:
Reference: https://github.com/moby/moby/issues/36562#issuecomment-372347544 

The named pipe in Windows is of message type and the handle to the pipe is in message read mode. Removing the check to allow for Windows named pipes to work.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -181,9 +181,12 @@ func DialPipe(path string, timeout *time.Duration) (net.Conn, error) {
 		return nil, err
 	}
 
-	if state&cPIPE_READMODE_MESSAGE != 0 {
+	/**
+	Windows support message type pipes in message-read mode only. Removing this check to allow for windows named pipes.
+	*/
+	/*if state&cPIPE_READMODE_MESSAGE != 0 {
 		return nil, &os.PathError{Op: "open", Path: path, Err: errors.New("message readmode pipes not supported")}
-	}
+	}*/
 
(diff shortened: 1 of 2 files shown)


mnltejaswini:
Also, we will treat ERROR_MORE_DATA as success and return the remaining number of bytes to be read.

pipe.go
```diff
@@ -181,9 +181,12 @@ func DialPipe(path string, timeout *time.Duration) (net.Conn, error) {
 		return nil, err
 	}
 
-	if state&cPIPE_READMODE_MESSAGE != 0 {
+	/**
+	Windows support message type pipes in message-read mode only. Removing this check to allow for windows named pipes.
```
thaJeztah: Looks like this message could be better put as a commit message, and the code removed?

andrewhsu:
@jstarks ptal

johnstep:
This is what I mentioned back in March, https://github.com/moby/moby/issues/36562#issuecomment-377583510, but I never heard back. Is this really the correct fix? Why is the VMbus proxy pipe (assuming that is how it works) a different mode?

johnstep:
This is what I mentioned back in March, https://github.com/moby/moby/issues/36562#issuecomment-377583510, but I never heard back. Is this really the correct fix? Why is the VMBus proxy pipe (assuming that is how it works) a different mode?

PR #79, opened 2018-06-06
Remove CRLF

mmatur:
Hi,

As we commit vendor, the CRLF breaks our dependencies validation system.


jhowardmsft:
Yup, should be complete now.

mmatur:
Hi @jterry75 @jhowardmsft,

Could you explain me why you have close this PR?

Line separators of these 3 files are different than others files. It can cause issue when dependencies are vendoring like we do in [Træfik](https://github.com/containous/traefik)

jhowardmsft:
@mmatur - ah, I didn't see this PR, but fixed by https://github.com/Microsoft/go-winio/pull/92

Issue #46, opened 2017-04-09
Integrate with the Go runtime’s event loop


DemiMarie:
This is a feature request for integration with the Go runtime’s event loop.  This would mean that operations appeared to block, but in fact did not, just like with the rest of Go.

tmm1:
I believe this is already the case.. all the APIs in this package block, and use channels under the hood to communicate with the completion port. The wrappers provided even conform to common golang interfaces like `io.Reader` and `io.Writer`.

gdamore:
My pull request #80 eliminates the use of waitNamedPipe, and so consequently should completely address this issue -- all the other uses in the code use IO completion ports sensibly to play well with the go runtime.

PR #80, opened 2018-06-24
fixes #67 DialPipe problem with multiple calls / waiting for busy pipe

### LINKED ISSUE #67: DialPipe problem with multiple calls / waiting for busy pipe

gdamore:
(I believe this change also should be used instead of PR #75 -- that PR is potentially extraordinarily buggy.)

This changes a few things to try to ensure that we never wind up with
a result of ERROR_FILE_NOT_FOUND, due to a race between closing the
last pipe instance and opening the next.

First we keep an "open" client instance (unused) while the listener
is open, so that we are guaranteed to always have an active pipe
instance.  This means attempts to open while no other instances exist
result in ERROR_PIPE_BUSY instead of ERROR_FILE_NOT_FOUND.

Second we have changed the loop for dialing to eliminate a race condition
that is more or less inherent in WaitNamedPipe when synchronizing with
CreateFile.  The real timeout needs to be some larger value than the
WaitNamedPipe timeout, and furthermore WaitNamedPipe is not very nice
with the Go runtime, since it is a blocking system call.  Instead we
just put the goroutine to sleep for 10 milliseconds, and keep retrying
the CreateFile until the maximum timeout is reached.  If no timeout is
specified we assume a reasonable and large default of 5 seconds, which is
similar to a TCP connection timeout.

This isn't perfect, as a client attempting to connect to an extremely
busy pipe server can be starved out by other clients coming in while
it is in that brief sleep, but this potential race was already present
with WaitNamedPipe.  The numerous retries (by default 500 retries!)
mean its pretty unlikely to occur, and if a single client hits the
race once, it has an excellent chance of getting in the next cycle.

(A real "fix" that is completely race free and fair would require
changes in the underlying Named Pipe implementation, or some other
kind of external coordination.)

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -140,7 +140,7 @@ func DialPipe(path string, timeout *time.Duration) (net.Conn, error) {
 	if timeout != nil {
 		absTimeout = time.Now().Add(*timeout)
 	} else {
-		absTimeout = time.Now().Add(time.Second * 5)
+		absTimeout = time.Now().Add(time.Second * 2)
 	}
 	var err error
 	var h syscall.Handle
@@ -200,7 +200,6 @@ type acceptResponse struct {
 
 type win32PipeListener struct {
 	firstHandle        syscall.Handle
-	clientHandle       syscall.Handle
 	path               string
 	securityDescriptor []byte
 	config             PipeConfig
@@ -303,8 +302,6 @@ func (l *win32PipeListener) listenerRoutine() {
 	}
 	syscall.Close(l.firstHandle)
 	l.firstHandle = 0
-	syscall.Close(l.clientHandle)
-	l.clientHandle = 0
 	// Notify Close() and Accept() callers that the handle has been closed.
 	close(l.doneCh)
 }
@@ -362,9 +359,13 @@ func ListenPipe(path string, c *PipeConfig) (net.Listener, error) {
 		syscall.Close(h)
 		return nil, err
 	}
+	// Close the client handle. The server side of the instance will
+	// still be busy, leading to ERROR_PIPE_BUSY instead of
+	// ERROR_NOT_FOUND, as long as we don't close the server handle,
+	// or disconnect the client with DisconnectNamedPipe.
+	syscall.Close(h2)
 	l := &win32PipeListener{
 		firstHandle:        h,
-		clientHandle:       h2,
 		path:               path,
 		securityDescriptor: sd,
 		config:             *c,
--- a/pipe.go
+++ b/pipe.go
@@ -15,7 +15,6 @@ import (
 //sys connectNamedPipe(pipe syscall.Handle, o *syscall.Overlapped) (err error) = ConnectNamedPipe
 //sys createNamedPipe(name string, flags uint32, pipeMode uint32, maxInstances uint32, outSize uint32, inSize uint32, defaultTimeout uint32, sa *syscall.SecurityAttributes) (handle syscall.Handle, err error)  [failretval==syscall.InvalidHandle] = CreateNamedPipeW
 //sys createFile(name string, access uint32, mode uint32, sa *syscall.SecurityAttributes, createmode uint32, attrs uint32, templatefile syscall.Handle) (handle syscall.Handle, err error) [failretval==syscall.InvalidHandle] = CreateFileW
-//sys waitNamedPipe(name string, timeout uint32) (err error) = WaitNamedPipeW
 //sys getNamedPipeInfo(pipe syscall.Handle, flags *uint32, outSize *uint32, inSize *uint32, maxInstances *uint32) (err error) = GetNamedPipeInfo
 //sys getNamedPipeHandleState(pipe syscall.Handle, state *uint32, curInstances *uint32, maxCollectionCount *uint32, collectDataTimeout *uint32, userName *uint16, maxUserNameSize uint32) (err error) = GetNamedPipeHandleStateW
 //sys localAlloc(uFlags uint32, length uint32) (ptr uintptr) = LocalAlloc
@@ -134,12 +133,14 @@ func (s pipeAddress) String() string {
 }
 
 // DialPipe connects to a named pipe by path, timing out if the connection
-// takes longer than the specified duration. If timeout is nil, then the timeout
-// is the default timeout established by the pipe server.
+// takes longer than the specified duration. If timeout is nil, then we use
+// a default timeout of 5 seconds.  (We do not use WaitNamedPipe.)
 func DialPipe(path string, timeout *time.Duration) (net.Conn, error) {
 	var absTimeout time.Time
 	if timeout != nil {
 		absTimeout = time.Now().Add(*timeout)
+	} else {
+		absTimeout = time.Now().Add(time.Second * 5)
 	}
 	var err error
 	var h syscall.Handle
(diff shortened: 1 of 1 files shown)


gdamore:
Hmm... I probably would have liked to have added my own copyright and update to the code... (still MIT licensed.) If its possible to do that somewhere, text more or less to the effect of:

Portions Copyright 2018 Garrett D'Amore <garrett@damore.org>

(or skip the Portions word.) . If this is too much a hassle, then don't worry about it.

gdamore:
This probably also fixes #46 -- since WaitNamedPipe was the only blocking code remaining.

carlfischer1:
cc @jstarks @johnstep

olljanat:
FYI. I was able to reproduce original issue by running my test version of Portainer on Windows Server, version 1803 and I cannot see it anymore on version which contains content from this PR (more detail on that portainer PR).

Any change to get this one merged?

StefanScherer:
I can confirm that this PR helps Portainer work stable with the named pipe bind mounted into a Windows container running a Docker swarm on Windows Server 1803 with Docker EE 18.03.1-ee-1

gdamore:
@jstarks @johnstep Ping?  This PR addresses a very important reliability issue, and at least three different projects are waiting for this to be integrated.  If there is something wrong with this, or concerns with what I've done here, then feedback would very much be appreciated.

johnstep:
I do not have permission to merge this.

aiminickwong:
@jstarks   time to allow merge this pull ?

gdamore:
Hard to believe I submitted this 19 days ago; the silence from the repo maintainers is deafening.

jstarks:
Sorry for the delays. I'm looking at this now. I want to write a quick test to understand the behavior of something before I merge this.

jstarks:
OK, I think this pull request is actually three changes:
1. Increase the default timeout to 5 seconds.
2. Poll instead of using WaitNamedPipe.
3. Keep the original dummy client handle open for the lifetime of the pipe server.

I understand why increasing the timeout may be useful, but this seems like something the client can easily do without a change to go-winio. I do suspect that the default 50ms is probably too short to be useful given the problems with fairness. But I'm worried that extending it to a full 5 seconds might affect existing clients negatively. Maybe a compromise would be something like 250ms.

WaitNamedPipe has its problems. As you mention, it blocks the OS thread (which go-winio generally tries to avoid), it has various races, it does not guarantee that the client will win the race and be able to connect to the pipe, and most importantly for #67, it does not work reliably inside containers. It does have the advantage, though, that it wakes up immediately when a connection is available. I wonder if a better change wouldn't be to keep the WaitNamedPipe call, but if it fails any reason other than timeout, to sleep for 10ms and loop around again. That would probably resolve #67 and other race conditions.

This suggestion doesn't fix the issue that WaitNamedPipe blocks the OS thread, of course. If this is important to resolve, I would suggest making a change to use `FSCTL_PIPE_WAIT` with an asynchronous DeviceIoControl call. This fsctl is what WaitNamedPipe uses, and it is documented as part of the SMB protocol. It appears to have a non-blocking implementation.

I don't understand the third change. I can't see any behavior differences whether I keep the original client handle open or not. Unless you are aware of a behavior that I am not, I think it's better to close that client handle to avoid unnecessary resource consumption.

olljanat:
@jstarks thanks for good comments.

I changed default timeout to 250 ms and added client handle closing now on this commit: https://github.com/olljanat/go-winio/commit/972aaec17501edc2ae66b43541f799eec50cf7c5 and did some testing with Portainer and I can tell that this combination at least still fixes #67

@gdamore I think that it is better if you comment about suggestion to still use WaitNamedPipe?

gdamore:
I think using WaitNamedPipe is probably more trouble than it's worth.  In
theory it would be better with the immediate wake up but as it doesn't
actually correct anything and as it blocks the caller I think it's better
not to use it and aim for simplicity.

5s was chosen to match TCP timeouts making this behave more like other
net.Pipe implementations. That said it is trivial to override as that is
just a default so I am happy with whatever you think is best there.

Keeping the old pipe open was intended to make it easier to distinguish
between the two cases of a listener not existing at all and a listener that
has not yet accepted the connection.  The former case you want to fail fast
without a time out.  The latter should keep trying because presumably
*something* is there and you have just missed it.

Again the semantic I am trying to provide is more like TCP or UNIX domain
pipes which both behave this way.

olljanat:
@jstarks / @gdamore so what we can do get this one fixed?

There is already three PRs #75 , this one and #84 which all are trying to solve same issue but bit different way. We should choose which one to focus and do needed changes to get is merged.

gdamore:
Hmm.. I have a question.

If the client handle is closed, but we haven't called DisconnectNamedPipe nor closed the server handle, is the server pipe instance still retained and busy (so no new client handles will connect to it, and so that the we won't get ERROR_NOT_FOUND in CreateFile on the client?) 

If so, then closing the file handle would be perfectly reasonable.  In such a case, there would be no need to retain the client handle.  A better comment explaining this would be helpful.  There is still that race condition I mentioned, where some other clients connects before we do.  *That* would be unfortunate, and again would be easily fixed in a follow up.

jstarks:
What I've observed (and I think we have tests to confirm) is that keeping the server handle open is sufficient to retain ownership of the pipe and to ensure that clients get ERROR_PIPE_BUSY. So keeping the client handle open is not necessary.

gdamore:
Do you want me to follow up with a modification to this PR for the above alterations (2 second timeout, close client handle) or address at your end?

gdamore:
Ok, coming shortly.

gdamore:
Done.  Would be good to test it before integrating. :-)

jstarks:
Thanks! I'll take a look a little later today and merge it. If @olljanat has time to validate it in his workload, that would be useful.

olljanat:
@jstarks my workload looks working very nicely with this one.
@gdamore good work 👍

jstarks:
Thanks @gdamore for the fix, @olljanat for verification, and everyone for their patience on this.

PR #82, opened 2018-06-25
Support pipe message read mode

jstarks:
Message read mode is the only read mode supported for vSMB named pipes, which are used to provide named pipe mapping for Hyper-V-isolated Windows containers. Transparently support this mode.


jstarks:
This supersedes #80.

gdamore:
I like this approach.  Now I think maybe I should take care to ensure that message read mode is supported in my C code that deals with named pipes .

thaJeztah:
@jstarks is it possible to tag a new release?

Issue #83, opened 2018-07-14
Race condition in construction of `win32PipeListener`


djsweet:
During the construction of `win32PipeListener`, an unused named pipe instance gets created. The `ListenPipe` constructor attempts to block this unused instance with a `CreateFile` call [on line 354](https://github.com/Microsoft/go-winio/blob/67921128fb397dd80339870d2193d6b1e6856fd4/pipe.go#L354), but this cannot happen transactionally during the creation of the named pipe: between line 349, just after the instance is created, and line 354, an optimistic named pipe client can connect to the pipe instance before the listener itself has a chance to block it. This causes listener construction to fail, despite the existence of a valid client on the named pipe handle.

I have a general use case where this race condition cannot be worked around: the named pipe client starts the named pipe server, and uses the existence of the pipe name to assume that the server is ready to accept connections. I suspect that #67 may be caused by this race condition as well.

PR #84, opened 2018-07-14
Eliminate initial connection race condition in win32PipeListener (addresses #83)

djsweet:
Previously, win32PipeListener would create an initial pipe instance,
open a client connection on it, then immediately close that connection.
The reasoning behind this was unclear, but it appears to have been a
mechanism for claiming the pipe name in the pipe namespace. If the
listener-side single client could not be created, the listener would
abort. This created a race condition where a legitimate client could
attach itself to the named pipe before the listener could create its own
client, causing the listener to incorrectly abort due to a legitimate
client connection.

This commit alters the process for listening for and accepting new
connections: when the win32PipeListener is constructed, the first
pipe is still created as before, but it is now also used as the first
client connection as well. In order to maintain the namespace stake,
the listenerRoutine() creates another named pipe instance before
returning the connected instance to a pending Accept(), so that there is
at least one pending named pipe instance at all times.

There are some caveats to this approach: transient errors from
CreateNamedPipe will not be returned until the _next_ Accept() call, so
there may be instances where calling clients may correct the cause
of the error, but still receive the same error across two Accept()
calls. ~~Additionally, if transient errors from CreateNamedPipe begin
occurring and all other named pipe instances are closed, the listener
loses its namespace stake and _may_ accidentally clobber other processes
if the transient errors resolve themselves.~~

~~It may be worth investingating closing the listener "from the inside"
if transient errors begin to occur in CreateNamedPipe, so that the
calling client is forced to create a new Listener and detect whether
its stake in the pipe namespace has been lost.~~

EDIT: As of commit [1c4c364](https://github.com/Microsoft/go-winio/pull/84/commits/1c4c364c32423537103d015142d80a61c6edb0cf), the namespace stake is guaranteed by having instances of `win32Pipe` "donate" their open instances back to the listener if the listener could not create a new pipe instance. This ensures that at least one open pipe instance is bound to the expected name through the lifetime of the listener, which also ensures that the stake in the namespace is not lost. It is currently using `sync.Mutex` rather heavily, but it might be possible to replace that with `atomic.CompareAndSwapPointer` instead.

Commit [1c4c364](https://github.com/Microsoft/go-winio/pull/84/commits/1c4c364c32423537103d015142d80a61c6edb0cf) also contains a fix for #85. This might be worth breaking out into a new PR.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -7,6 +7,8 @@ import (
 	"io"
 	"net"
 	"os"
+	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 	"unsafe"
@@ -55,6 +57,12 @@ var (
 type win32Pipe struct {
 	*win32File
 	path string
+	// If this instance of a pipe was created by a listener, the Close()
+	// method may attempt to return its instance to the listener to be
+	// re-used iff the listener does not already have another instance
+	// prepared for connection, usually as the consequence of an error
+	// returned by createNamedPipe.
+	listener *win32PipeListener
 }
 
 type win32MessageBytePipe struct {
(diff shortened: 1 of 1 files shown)


djsweet:
I have a separate branch in which I've been trying out `atomic.CompareAndSwapPointer` instead of mutexes at [https://github.com/djsweet/go-winio/tree/pipe-listener-no-mutex], and am currently using it for my own purposes. However, I'm not terribly eager to integrate that in, because the Named Pipe API is quite inconsistent when it comes to the myriad ways of closing a connection.

That branch had to treat `ERROR_PIPE_NOT_CONNECTED` as `io.EOF` for compatibility with existing consumers and consistency with the `net.Conn` interface. However, this was not an enhancement but a requirement: client handles to named pipes closed with `DisconnectNamedPipe` return `ERROR_PIPE_NOT_CONNECTED` instead of the expected `ERROR_HANDLE_EOF`. This has negative implications for clients of servers consuming this library. This current PR also might use `DisconnectNamedPipe`, but not as frequently as the other branch, which always uses `DisconnectNamedPipe`.

As a result, replacing the mutex usage will take some thought and create incompatibilities that we probably don't want, but I'm not sure if they're avoidable while fixing the initial listener race.

gdamore:
The Named Pipe API is quite inconsistent.  Full stop.

Thanks for the heads up about ERROR_PIPE_NOT_CONNECTED.

ERROR_WINDOWS_HAS_TOO_MANY_ERROR_CODES

tankbusta:
I'm seeing some odd behavior that I think might be related to this? My listener doesn't seem to accept the first connection to the named pipe. If I restart my client, everything works and every connection is accepted as expected.

jterry75:
@jstarks - PTAL

tankbusta:
Yep @djsweet's PR is fixing the issue for me. I was curious and tried running all the pipe tests from `master` and the vast majority of them are failing.

PR #88, opened 2018-08-02
Update FileAttributes size to work across all platforms

jiria:
FileAttributes is expected to be 64 bit wide when calling into Windows APIs, no matter the platform. This fix is ensuring that.

### DIFF
--- a/backuptar/tar.go
+++ b/backuptar/tar.go
@@ -303,7 +303,7 @@ func FileInfoFromHeader(hdr *tar.Header) (name string, size int64, fileInfo *win
 		if err != nil {
 			return "", 0, nil, err
 		}
-		fileInfo.FileAttributes = attr
+		fileInfo.FileAttributes = uint32(attr)
 	} else {
 		if hdr.Typeflag == tar.TypeDir {
 			fileInfo.FileAttributes |= syscall.FILE_ATTRIBUTE_DIRECTORY
--- a/backuptar/tar.go
+++ b/backuptar/tar.go
@@ -303,7 +303,7 @@ func FileInfoFromHeader(hdr *tar.Header) (name string, size int64, fileInfo *win
 		if err != nil {
 			return "", 0, nil, err
 		}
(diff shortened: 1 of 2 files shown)


jhowardmsft:
LGTM. @jterry75 Could you also take a look while @jstarks is out of town.

jterry75:
@jiria - Are you sure? [_FILE_BASIC_INFO](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/ns-winbase-_file_basic_info). Seems to be that its a `DWORD` in the typedef.

jiria:
@jterry75 Good catch, made me question my sanity. The pre-existing comment in the file helped. Structures are padded at the end so that the size of the structure is divisible by the size of the largest element. So in the case of 64 bit platform, sizeof(uintptr) was 8, so no extra padding was needed. On 32 bit platform, sizeof(uintptr) or sizeof(DWORD) are 4, so there is invisible 4 bytes of alignment. Since go does not observe these, the buffer size was only 36, not 40. 

Having said that, I do not think this change is the best way to go about fixing this. We could add 4 bytes of dummy padding into the GO structure, the Windows code seems to be testing only the minimum size of the buffer, so it would not break 64 bit platforms. Or do you have some other thoughts on how to amend this?

jterry75:
The `cgo` standard would be to add a `uint32 pad` variable after the `DWORD` I believe. I'm actually surprised that this works at all I would of thought it would write the attribute data into the high 4 bytes and your comparisons to the actual values would be wrong. Either way seems right to just make it say
```go
FileAttributes uint32;
pad uint32
```

jterry75:
LGTM. Did your tests with this work as expected?

jiria:
Yes, I had mirrored these changes on top of fork of moby/moby and I can build dockerd and run it successfully (tested docker import, docker run). Please let me know if there is more tests that you would want me to run. Other than that, I think we can complete this.

PR #91, opened 2018-08-23
Added DetachVhd

jhowardmsft:


jhowardmsft:
@jterry75 Looks better now the line endings are fixed. Adds a `DetachVhd` entrypoint. I'm intending using this in docker to resolve locked files from when the daemon is terminated and leaves VHDs mounted which can't be removed when a new daemon is started.

PR #94, opened 2018-09-13
treat ErrFileClosed as EOF on Read()

zanderz:
An async waiting read can return ErrFileClosed if the pipe is closed on another gofunc. This changes it to be considered an EOF condition instead.
@maxtaco @akalin


maxtaco:
can this patch be upstreamed so we don't have to maintain this fork?

Issue #26, opened 2016-06-07
panic:Failed to find SetFileCompletionNotificationModes procedure in kernel32.dll: The specified procedure could not b e found.


GameXG:
win 2008 before the version does not have this function.
Hope to solve this compatibility problem.

PR #101, opened 2018-10-30
avoid SetFileCompletionNotificationModes crash

zanderz:
According to Microsoft's own documentation, there is a known bug where SetFileCompletionNotificationModes crashes on some systems (see https://support.microsoft.com/kb/2568167 for details) and a user of ours has experienced this with our software. This adapts the precautions added to Golang itself (https://go-review.googlesource.com/c/go/+/69870/).
@maxtaco

### DIFF
--- a/file.go
+++ b/file.go
@@ -10,12 +10,55 @@ import (
 	"sync/atomic"
 	"syscall"
 	"time"
+	"unsafe"
 )
 
 //sys cancelIoEx(file syscall.Handle, o *syscall.Overlapped) (err error) = CancelIoEx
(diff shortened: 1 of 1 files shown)


runner-mei:
any progress

TBBle:
Does this also replace #27, fixing #26? If I'm reading [the Golang-side change](https://go.googlesource.com/go/+/c37647fc3abd3aa9307d5abbce63c9d6c7c70602%5E%21/), it also handles Windows XP, so I expect the answer is "yes". I'm not sure _how_ it handles XP though.

I'm also curious why the second hunk from #27 wasn't needed here, the change in `asyncIo`, performing the result-wait even if err was not `syscall.ERROR_IO_PENDING`.

PR #103, opened 2018-12-13
Add basic Go ETW implementation

kevpar:
I'm still working on a few things, but sending this out so others can take a look.


etw/event.go
```diff
@@ -0,0 +1,87 @@
...
+func NewEvent(name string, descriptor *EventDescriptor) *Event {
+	return &Event{
+		Descriptor: descriptor,
+		Metadata:   NewEventMetadata(name),
+		Data:       &EventData{},
+	}
+}
+
+// EventDescriptor represents various metadata for an ETW event.
+type EventDescriptor struct {
```
jterry75: Since you broke the others up do you want `EventDescriptor` in its own file as well?

etw/event.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// NewEventDescriptor returns an EventDescriptor initialized for use with
+// tracelogging.
+func NewEventDescriptor() *EventDescriptor {
+	return &EventDescriptor{
+		id:      0,
+		version: 0,
+		Channel: ChannelTracelogging,
+		Level:   LevelVerbose,
```
jterry75: Seems like the wrong default. Should we add a `NewEventDescriptor(level)` instead?

etw/etw.go
```diff
@@ -0,0 +1,7 @@
+package etw
```
jterry75: This really should be in `pkg/etw` to align with golang recommendations

jterry75: Yes I know none of the other packages in this repo adhere to this :)

etw/eventdata.go
```diff
@@ -0,0 +1,23 @@
...
+
+// EventData maintains a buffer which builds up the data for an ETW event. It
+// needs to be paired with EventMetadata which describes the event.
+type EventData struct {
+	buffer bytes.Buffer
+}
+
+// NewEventData returns a new EventData with an empty buffer.
+func NewEventData() *EventData {
+	return &EventData{}
```
jterry75: NIT: If we don't initialize any fields I prefer the syntax `new(EventData)` but certainly not required.

etw/eventdata.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+// AddString appends the data for a string to the end of the buffer.
+func (ed *EventData) AddString(data string) {
```
jterry75: Do you plan on adding an overload for all types? Should we just use `AddItem(data interface{})`

etw/eventmetadata.go
```diff
@@ -0,0 +1,51 @@
...
+	"bytes"
+	"encoding/binary"
+)
+
+// InType indicates the type of data contained in the ETW event.
+type InType byte
+
+// Various InType definitions for tracelogging.
+const (
+	InTypeNull InType = iota
```
jterry75: Are these made up? Can we use any types here or do we have to use exactly these types?

etw/eventmetadata.go
```diff
@@ -0,0 +1,51 @@
...
+// event. It needs to be paired with EventData which describes the event.
+type EventMetadata struct {
+	buffer bytes.Buffer
+}
+
+// NewEventMetadata returns a new EventMetadata with event name and initial
+// metadata written to the buffer.
+func NewEventMetadata(name string) *EventMetadata {
+	em := EventMetadata{}
+	binary.Write(&em.buffer, binary.LittleEndian, uint16(0))    // Length placeholder
```
jterry75: Do you really need the `&em.buffer` seems like since `bytes.Buffer` is `io.Writer` you should be able to just pass `em.buffer`.

etw/logrus/hook.go
```diff
@@ -0,0 +1,78 @@
+package hook
```
jterry75: Is a nested package really the right approach here? Why not just have this in the etw package?

etw/logrus/hook.go
```diff
@@ -0,0 +1,78 @@
...
+	"golang.org/x/sys/windows"
+)
+
+// Hook is a Logrus hook which logs received events to ETW.
+type Hook struct {
+	provider *etw.Provider
+}
+
+// NewHook registers a new ETW provider and returns a hook to log from it.
+func NewHook(providerName string, providerID *windows.GUID) (*Hook, error) {
```
jterry75: `*windows.GUID`? Why `*` isn't it a `[16]byte`?

etw/logrus/hook.go
```diff
@@ -0,0 +1,78 @@
...
+	descriptor.Level = etw.Level(e.Level)
+
+	event := etw.NewEvent("LogrusEntry", descriptor)
+
+	event.Metadata.AddField("Message", etw.InTypeAnsiString)
+	event.Data.AddString(e.Message)
+
+	for k, v := range e.Data {
+		switch reflect.TypeOf(v).Kind() {
+		case reflect.String:
```
jterry75: Can we at least handle all the `InType` in the list before using a generic format string?

jterry75: Also do a type switch rather than this reflection.

etw/sample/sample.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"golang.org/x/sys/windows"
+)
+
+func callback(sourceID *windows.GUID, isEnabled uint32, level byte, matchAnyKeyword uint64, matchAllKeyword uint64, filterData uintptr) {
+	fmt.Printf("Callback: isEnabled=%d, level=%d, matchAnyKeyword=%d\n", isEnabled, level, matchAnyKeyword)
+}
+
+func main() {
+	providerID := windows.GUID{0xdd2062c6, 0x5d1b, 0x4a0f, [8]uint8{0xbd, 0xb9, 0x22, 0x28, 0xbc, 0xb1, 0x07, 0x7c}}
```
jterry75: What GUID is this? Can you give a comment here?

etw/sample/sample.go
```diff
@@ -0,0 +1,53 @@
...
+			logrus.Error(err)
+		}
+	}()
+
+	reader := bufio.NewReader(os.Stdin)
+
+	fmt.Println("Press enter to log an event")
+	reader.ReadString('\n')
+
+	event := etw.NewEvent("TestEvent", etw.NewEventDescriptor())
```
jterry75: These could really be test cases. I think it would be nice to show a test case for the various `InTypes` which would make this more clear how to handle it. My guess is that zero consumers will write this code. Instead they will write.

```go
logrus.WithFields(logrus.Fields{
    "id", idString,
    "value", valueInt32,
    "time", time.Now(),
}.Debug("Trace Point")
```
And what we want to see is how the hook writes a "String, int32, time -> int?" etc.

etw/event.go
kevpar: I mostly moved EventData and EventMetadata to their own files because I anticipate their set of functions growing over time as we support for data types. EventDescriptor should stay small, so I'm okay with it staying in this file.

etw/event.go
kevpar: For tracelogging events (which is what we are supporting here), the default is actually to use verbose for level, and in many cases it's not changed. I'll add a comment clarifying that these are the   standard defaults for tracelogging.

etw/eventdata.go
kevpar: I mostly put this in for symmetry with how EventMetadata is initialized, and in case we need to do any work at init in the future. I'm not sure those are fantastic reasons, though. :)

etw/eventdata.go
kevpar: It will depend on the specifics of other types we want to support, and how tracelogging requires them to be packed. I don't think we can support a generic AddItem() at this level, but rather at a higher level which calls into these functions to do the work. Right now i"m planning to add more here as needed while testing this out.

As an example of another way to pack data: tracelogging supports counted strings, which are prefixed with length.

etw/eventmetadata.go
kevpar: These have to match the tracelogging definitions (can be found in TraceLoggingProvider.h), so that decoding properly works. There are more than I put here, I just threw some in as a starting point.

etw/eventdata.go
kevpar: For a lot of blittable types like numbers we should be able to write their data using a single function.

etw/eventmetadata.go
kevpar: I tried that, but Go doesn't like it:

```
cannot use em.buffer (type bytes.Buffer) as type io.Writer in argument to binary.Write:
        bytes.Buffer does not implement io.Writer (Write method has pointer receiver)
```

Sounds like there's a `func (*bytes.Buffer) Write` but not a `func (bytes.Buffer) Write`?

etw/logrus/hook.go
kevpar: I wanted to have the etw package just be the base support for ETW itself, and put things building on top of it in other packages. Seemed like good separation, and ensures the etw package exposes the right set of things to allow something meaningful to be built with it.

etw/logrus/hook.go
kevpar: windows.GUID is defined as this, which matches the GUID struct in Windows C/C++ code:

```
type GUID struct {
	Data1 uint32
	Data2 uint16
	Data3 uint16
	Data4 [8]byte
}
```

I used a pointer because that's how I usually see GUIDs passed around in C/C++. My assumption was the cost is about the same there as in Go, so making the same decision seemed reasonable. Happy to change that though. I don't have a good heuristic for when Go code tends to use pointers or not to pass structs around.

etw/sample/sample.go
kevpar: Every provider needs to define its own GUID which uniquely identifies it to ETW. I'll add a comment explaining this.

Also, tracelogging has an algorithm to generate the provider GUID based on the provider name. Once I get that implemented, we can just pass a name in most cases, and not worry about the GUID here.

etw/sample/sample.go
kevpar: I can move this to be test code instead of a sample. I'm not sure there's anything meaningful to check against to see if the test succeeds (other than validating an error didn't get returned), though.

etw/logrus/hook.go
kevpar: Will do.

etw/etw.go
kevpar: ETW can be the trailblazer :)

etw/eventdata.go
jterry75: Eh, whatever. `new` is mostly historic and sync this is a struct we can use the `&` syntax. Just leave it.

etw/eventdata.go
jterry75: Should we just add a function for each intrinsic type at least then?

etw/eventmetadata.go
jterry75: Can you write that in the comment. `These must match the values found in TraceLoggingProvider.h in the Windows SDK`. or something?

etw/eventmetadata.go
jterry75: Oh I see. No worries then. The compiler in my head made a mistake :)

etw/logrus/hook.go
jterry75: I guess what I am saying is that people don't write packages in go like:
```
pkg\datatypes
pkg\datatypes\mytype.go
pkg\datatypes\json
pkg\datatypes\json\mytype_json.go
pkg\datatypes\toml
pkg\datatypes\toml\mytype_toml.go
```
And in the `json` and `toml` files add the serializers.

The would write
```
pkg\datatypes
pkg\datatypes\mytype.go
pkg\datatypes\mytype_json.go
pkg\datatypes\mytype_toml.go
```
And then the caller can decide which ones to include.

So for this it seems:
```
pkg\etw
pkg\etw\stuff.go
pkg\etw\logrus_hook.go
pkg\etw\myother_hook.go
```
OR Two packages
```
pkg\etw
pkg\etwlogrus
```

etw/logrus/hook.go
jterry75: Oh wow. I didn't realize `x/sys/windows` had yet another `GUID` structure. Should we take a different public structure so that we don't have to force every caller to use this and include the `x/sys/windows` package to do so?

etw/sample/sample.go
jterry75: See my other comment about removing GUID. We should do that instead of requiring people to pass a `GUID` here.

etw/sample/sample.go
jterry75: I think we need to create a test listener so that we can verify that each event we generate is written as we expect into the platform. Would be really nice to know this code passes all tests. If we need to fake this we can at the syscall layer but I thought we could easily make a listener and just exercise the product in the test.

etw/logrus/hook.go
kevpar: If I go with the first approach, they'll have to all be in the same package, right? I'd rather ensure that the ETW Logrus hook can't use non-exported things from ETW, so I think they should be in different packages.

The second approach looks fine to me, though.

etw/logrus/hook.go
kevpar: If I can get the GUID generation from provider name working, I think we can just remove the GUID parameter from NewProvider (and add something like NewProviderWithID that has it). Then the majority of consumers could get away with not needing `x/sys/windows` as they wouldn't need to pass a GUID.

etw/eventdata.go
kevpar: I'll expand on this in a future PR.

pkg/etw/event.go
```diff
@@ -0,0 +1,89 @@
...
+	LevelWarning
+	LevelInfo
+	LevelVerbose
+)
+
+// Event represents a single ETW event. It can have field metadata and data
+// added to it, and then be logged via a provider to actually send it to ETW.
+type Event struct {
+	Descriptor *EventDescriptor
+	Metadata   *EventMetadata
```
jstarks: Is pointer to member necessary on all of these? Wondering if we can cut down on the number of allocations per event.

Also wondering if Descriptor needs to be public (and same for Metadata and Data pending discussion on my other comment).

pkg/etw/event.go
```diff
@@ -0,0 +1,89 @@
...
+		Keyword: 0,
+	}
+}
+
+// Identity returns the identity of the event. If the identity is not 0, it
+// should uniquely identify the other event metadata (contained in
+// EventDescriptor, and field metadata). Only the lower 24 bits of this value
+// are relevant.
+func (ed *EventDescriptor) Identity() uint32 {
+	return (uint32(ed.version) << 16) & uint32(ed.id)
```
jstarks: Is this supposed to be
```suggestion
	return (uint32(ed.version) << 16) | uint32(ed.id)
```

pkg/etwlogrus/hook.go
```diff
@@ -0,0 +1,78 @@
...
+		logrus.WarnLevel,
+		logrus.ErrorLevel,
+		logrus.FatalLevel,
+		logrus.PanicLevel,
+	}
+}
+
+// Fire receives each Logrus entry as it is logged, and logs it to ETW.
+func (h *Hook) Fire(e *logrus.Entry) error {
+	descriptor := etw.NewEventDescriptor()
```
jstarks: Would be nice to skip everything if the provider is not enabled.

pkg/etw/provider.go
```diff
@@ -0,0 +1,107 @@
...
+
+	var metadataBuffer bytes.Buffer
+	binary.Write(&metadataBuffer, binary.LittleEndian, uint16(0))                       // Write empty size for buffer (to update later)
+	binary.Write(&metadataBuffer, binary.LittleEndian, []byte(name))                    // Provider name
+	binary.Write(&metadataBuffer, binary.LittleEndian, byte(0))                         // Null terminator for name
+	binary.LittleEndian.PutUint16(metadataBuffer.Bytes(), uint16(metadataBuffer.Len())) // Update the size at the beginning of the buffer
+
+	return &Provider{
+		handle:   providerHandle,
+		metadata: &metadataBuffer,
```
jstarks: May as well store this as []byte instead of binary.Bytes.

pkg/etw/provider.go
```diff
@@ -0,0 +1,107 @@
...
+// WriteEvent writes a single event to ETW, from this provider.
+func (provider *Provider) WriteEvent(event *Event) error {
+	// Finalize the event metadata buffer by filling in the buffer length at the
+	// beginning.
+	binary.LittleEndian.PutUint16(event.Metadata.buffer.Bytes(), uint16(event.Metadata.buffer.Len()))
+
+	var dataDescriptors [3]eventDataDescriptor
+	dataDescriptors[0].set(eventDataDescriptorTypeProviderMetadata, provider.metadata)
+	dataDescriptors[1].set(eventDataDescriptorTypeEventMetadata, &event.Metadata.buffer)
+	dataDescriptors[2].set(eventDataDescriptorTypeUserData, &event.Data.buffer)
```
jstarks: My understanding (from TraceLoggingProvider.h) is that you can pass as many data descriptors as you want. It might be interesting to look into using the strings in place instead of copying them into a buffer.

pkg/etwlogrus/hook.go
```diff
@@ -0,0 +1,78 @@
...
+
+	event := etw.NewEvent("LogrusEntry", descriptor)
+
+	event.Metadata.AddField("Message", etw.InTypeAnsiString)
+	event.Data.AddString(e.Message)
+
+	for k, v := range e.Data {
+		switch v := v.(type) {
+		case string:
+			event.Metadata.AddField(k, etw.InTypeAnsiString)
```
jstarks: Would be nice to specify out type of UTF-8 (TlgOutUTF8 in TraceLoggingProvider.h).

etw/sample/sample.go
jstarks: I'm not 100% sure on this. I remember TraceLoggingProvider.h making it easy to avoid passing a GUID, and then later that was removed/heavily discouraged. Before we automatically generate a GUID based on the name, we should understand why that was so.

pkg/etwlogrus/hook.go
jstarks: Looks like to specify UTF-8 you just need to set the high bit on the InType byte and then the next byte is the out type.

etw/sample/sample.go
kevpar: Got a response:

> I don’t know of any reason not to rely on auto-generation. It’s also publicly documented (as EtwGuid) and used in EventSource as the default GUID generator for providers.

jstarks: OK good deal.

pkg/etw/provider.go
jstarks: (This is just something to think about. Not sure if it's a practical improvement since you still have to put the string length or null terminator in there.)

pkg/etw/eventdata.go
```diff
@@ -0,0 +1,23 @@
...
+type EventData struct {
+	buffer bytes.Buffer
+}
+
+// NewEventData returns a new EventData with an empty buffer.
+func NewEventData() *EventData {
+	return &EventData{}
+}
+
+// AddString appends the data for a string to the end of the buffer.
```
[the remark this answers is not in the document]
kevpar: There are some more complicated types that require separately specifying data and metadata.

For instance, to write an array of ints, you only do one call to write an array metadata, but have several calls to write each piece of data. That's a simple example, but once you start having types like nested structs and arrays of structs, it's useful to keep data and metadata apart.

I'm not opposed to adding some sort of helper for common types, in addition to the lower-level interface. I have been planning on waiting a little to see how the use cases for this code develop before putting that in.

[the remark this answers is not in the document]
jstarks: I think the important feedback here is that since this is a public package for others to consume, you should focus on making the exposed interface ergonomic and supportable. You can always expose the lower-level interface later, but exposing it now might make it harder to change the underlying implementation later.

E.g. for your int array example, I would guess most (all?) callers would rather just pass an int slice and let the underlying code handle the loop. This is less flexible since it requires that the caller have all their data available in a contiguous slice, but it's what TraceLogging.h already offers and that seems to be enough for most of Windows.

pkg/etwlogrus/hook.go
kevpar: My understanding is Go string literals are UTF8, but you can't assume an arbitrary string is UTF8. So I assume we shouldn't mark all strings we get from Logrus as UTF8 in this case, right?

pkg/etw/provider.go
kevpar: I'll take a look at how TraceLoggingProvider.h uses data descriptors. It looked like TraceLoggingDynamic.h just used 3 descriptors like we do here.

pkg/etwlogrus/hook.go
kevpar: I've changed the usage in the Logrus hook to specify UTF8. As the public-facing ETW interface evolves, I'll keep in mind just exposing UTF8 strings.

pkg/etw/event.go
kevpar: I made them pointers so you could e.g. replace Metadata with a pre-constructed metadata object, without needing a full copy. I don't have a good sense of what Go devs consider too large to copy around. If we don't care about the copy overhead, it should be just fine to get rid of the pointers here.

pkg/etw/provider.go
```diff
@@ -0,0 +1,107 @@
...
+	dataType  eventDataDescriptorType
+	reserved1 uint8
+	reserved2 uint16
+}
+
+func (descriptor *eventDataDescriptor) set(dataType eventDataDescriptorType, buffer *bytes.Buffer) {
+	// Passing a pointer to Go-managed memory as part of a block of memory is
+	// risky since the GC doesn't know about it. If we find a better way to do
+	// this we should use it instead.
+	descriptor.ptr = uint64(uintptr(unsafe.Pointer(&buffer.Bytes()[0])))
```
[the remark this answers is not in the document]
kevpar: This looks like a good approach, I'll do this.

As an aside, are you familiar with runtime.KeepAlive? Given the current reality of a non-relocating GC, it seems like it may be sufficient to protect these objects from being freed while the syscall is in progress.

pkg/etw/eventdata.go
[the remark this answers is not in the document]
kevpar: I've done some work to figure out more of what the API shape should be. It still needs to be fleshed out a bit more, but the code in sample.go demonstrates the difference between the low-level and high-level API. I anticipate most uses being able to consume the high-level API, and have ported the Logrus hook to use it as well.

pkg/etw/event.go
kevpar: Event is gone completely now. I'm still a bit curious as to what a good breakpoint would be on object size to change from pass-by-value to pass-by-pointer. Also, do you know if arrays are allocated on the heap or not? That could make a big difference for bytes.Buffer since it contains a [64]byte array.

internal/etw/provider.go
```diff
@@ -0,0 +1,306 @@
...
+// https://blogs.msdn.microsoft.com/dcook/2015/09/08/etw-provider-names-and-guids/
+// The algorithm is roughly:
+// Hash = Sha1(namespace + arg.ToUpper().ToUtf16be())
+// Guid = Hash[0..15], with Hash[7] tweaked according to RFC 4122
+func providerIDFromName(name string) (*windows.GUID, error) {
+	namespace := []byte{0x48, 0x2C, 0x2D, 0xB2, 0xC3, 0x90, 0x47, 0xC8, 0x87, 0xF8, 0x1A, 0x15, 0xBF, 0xC1, 0x30, 0xFB}
+	buffer := &bytes.Buffer{}
+	buffer.Write(namespace)
+
+	nameUTF16, err := windows.UTF16FromString(strings.ToUpper(name))
```
jstarks: >UTF16FromString [](start = 27, length = 15)

utf16.Encode([]rune(s))

PR #104, opened 2019-01-07
Support ETW events with no data fields

kevpar:


kevpar:
@jterry75 PTAL

internal/etw/provider.go
```diff
@@ -202,7 +202,15 @@ func (provider *Provider) WriteEvent(name string, eventOpts []EventOpt, fieldOpt
 		opt(em, ed)
 	}
 
-	return provider.WriteEventRaw(descriptor, [][]byte{em.Bytes()}, [][]byte{ed.Bytes()})
+	// Don't pass a data blob if there is no event data. There will always be
+	// event metadata (e.g. for the name) so we don't need to do this check for
+	// the metadata.
+	dataBlobs := [][]byte{}
```
jterry75: Does this code pass tests if it instead reads
`var dataBlobs [][]byte`
I am wondering if the syntax `dataBlobs := [][]byte{}` initiates an alloc.

kevpar: I think it won't allocate. The standard says this about composite literals:

```A literal may omit the element list; such a literal evaluates to the zero value for its type.```

https://golang.org/ref/spec#Composite_literals

PR #105, opened 2019-01-08
Support ETW activity ID

kevpar:


kevpar:
@jterry75 PTAL

internal/etw/eventopt.go
```diff
@@ -12,23 +23,35 @@ func WithEventOpts(opts ...EventOpt) []EventOpt {
...
 // value (top 4 bits are ignored) which are interpreted by the event consumer.
 func WithTags(newTags uint32) EventOpt {
-	return func(descriptor *EventDescriptor, tags *uint32) {
-		*tags |= newTags
+	return func(options *eventOptions) {
+		options.tags |= newTags
+	}
+}
+
+func WithActivityID(activityID *windows.GUID) EventOpt {
```
jterry75: Does this have to be a GUID or can it be any field?

internal/etw/eventopt.go
```diff
@@ -1,9 +1,20 @@
 package etw
 
+import (
+	"golang.org/x/sys/windows"
+)
+
+type eventOptions struct {
+	descriptor        *EventDescriptor
```
jterry75: Doesn't really matter but since you never allow this to be `nil` you dont need it to be a `ptr`.

internal/etw/eventopt.go
kevpar: Activity ID must be a GUID, we pass it to the LPCGUID parameter on EventWriteTransfer: https://docs.microsoft.com/en-us/windows/desktop/api/evntprov/nf-evntprov-eventwritetransfer

internal/etw/eventopt.go
kevpar: Going to treat this as Won't Fix for now, but will keep in mind in the future.

kevpar:
Acknowledged about the -s, I'll do that next time.

PR #107, opened 2019-01-15
Add a simple tool to generate provider ID's

jterry75:
Signed-off-by: Justin Terry (VM) <juterry@microsoft.com>


tools/etw-provider-gen/main.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+	p, err := etw.NewProvider(*pn, nil)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to convert provider-name: '%s' with err: '%s", *pn, err)
+		os.Exit(1)
+	}
+	defer p.Close()
+
+	var resultp *uint16
+	err = stringFromCLSID(p.ID, &resultp)
```
kevpar: You could also use the guidToString function I wrote here: https://github.com/Microsoft/go-winio/blob/master/internal/etw/sample/sample.go#L21

jterry75: Deal

internal/etw/sample/sample.go
```diff
@@ -47,7 +29,7 @@ func main() {
 		}
 	}()
 
-	fmt.Println("Provider ID:", guidToString(provider.ID))
+	fmt.Printf("Provider ID: %s\n", provider.ID)
```
kevpar: I think you need to just pass `provider` here, instead of `provider.ID`

PR #108, opened 2019-01-17
Set ETW provider info so it's always treated as TraceLogging

kevpar:
It turns out ETW only knew to parse our events as TraceLogging previously because they were set to the known TraceLogging channel (11). If you changed the event channel the event was not parsed correctly. This change tells ETW (via EventSetInformation with EventProviderSetTraits) that the provider uses TraceLogging, so events can be parsed even if the channel is changed.

Signed-off-by: Kevin Parsons <kevpar@ntdev.microsoft.com>

### DIFF
--- a/internal/etw/eventopt.go
+++ b/internal/etw/eventopt.go
@@ -36,6 +36,12 @@ func WithKeyword(keyword uint64) EventOpt {
 	}
 }
 
+func WithChannel(channel Channel) EventOpt {
+	return func(options *eventOptions) {
+		options.descriptor.Channel = channel
+	}
+}
+
 // WithTags specifies the tags of the event to be written. Tags is a 28-bit
 // value (top 4 bits are ignored) which are interpreted by the event consumer.
 func WithTags(newTags uint32) EventOpt {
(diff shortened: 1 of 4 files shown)


internal/etw/provider.go
```diff
@@ -41,6 +42,15 @@ const (
 	ProviderStateCaptureState
 )
 
+type eventInfoClass uint32
+
+const (
+	eventProviderInfoClassBinaryTrackInfo eventInfoClass = iota
```
jterry75: NIT: `eventInfoClass` is the type so these should be `eventInfoClass...`

internal/etw/eventopt.go
```diff
@@ -36,6 +36,12 @@ func WithKeyword(keyword uint64) EventOpt {
 	}
 }
 
+func WithChannel(channel Channel) EventOpt {
+	return func(options *eventOptions) {
+		options.descriptor.Channel = channel
```
jterry75: Where do you even use this?

kevpar: We don't use it anywhere. I added it when testing changing channel, and decided to leave it in rather than delete the code once it was written.

PR #109, opened 2019-01-26
archive/tar: remove excessive []byte(s) conversion

Quasilyte:
`copy` permits using to mix `[]byte` and `string` arguments without
explicit conversion. I removed explicit conversion to make code simpler.

Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>


jstarks:
Thanks for the change. This code was forked from an older version of Go's archive/tar implementation, where the code came from. I would recommend that new code use the latest version of Go's archive/tar, since it now supports arbitrary PAX extensions. For that reason, I do not want to take any changes to this implementation, even though otherwise this change is a good one.

PR #116, opened 2019-03-20
Add GrantVmGroupAccess

jhowardmsft:

@jterry75 @jstarks PTAL.

### DIFF
--- a/pkg/security/grantvmgroupaccess.go
+++ b/pkg/security/grantvmgroupaccess.go
@@ -1,0 +1,224 @@
+package security
+
+import (
+	"os"
+	"syscall"
+	"unsafe"
+
+	"github.com/pkg/errors"
+)
+
+type (
+	accessMask          uint32
+	accessMode          uint32
+	desiredAccess       uint32
+	inheritMode         uint32
+	objectType          uint32
+	shareMode           uint32
+	securityInformation uint32
+	trusteeForm         uint32
+	trusteeType         uint32
+
+	explicitAccess struct {
+		accessPermissions accessMask
+		accessMode        accessMode
+		inheritance       inheritMode
+		trustee           trustee
+	}
+
+	trustee struct {
+		multipleTrustee          *trustee
+		multipleTrusteeOperation int32
+		trusteeForm              trusteeForm
+		trusteeType              trusteeType
+		name                     uintptr
+	}
+)
+
+const (
+	accessMaskDesiredPermission accessMask = 0x12019f
+
+	accessModeGrant accessMode = 1
+
+	desiredAccessReadControl desiredAccess = 0x20000
+	desiredAccessWriteDac    desiredAccess = 0x40000
+
+	gvmga = "GrantVmGroupAccess:"
+
+	inheritModeNoInheritance                  inheritMode = 0x0
+	inheritModeSubContainersAndObjectsInherit inheritMode = 0x3
+
+	objectTypeFileObject objectType = 0x1
+
+	securityInformationDACL securityInformation = 0x4
+
+	shareModeRead  shareMode = 0x1
+	shareModeWrite shareMode = 0x2
+
+	sidVmGroup                   = "S-1-5-83-1-3166535780-1122986932-343720105-43916321"
+	sidVmWorkerProcessCapability = "S-1-15-3-1024-2268835264-3721307629-241982045-173645152-1490879176-104643441-2915960892-1612460704"
+
+	trusteeFormIsSid trusteeForm = 0
+
+	trusteeTypeWellKnownGroup trusteeType = 5
+)
+
+// GrantVMGroupAccess sets the DACL for a specified file or directory to
+// include Grant ACE entries for both the VM Group SID, and the VM Worker Process
+// Capability SID. This is a golang re-implementation of the same function in
+// vmcompute, just not exported in RS5. Which kind of sucks. Sucks a lot :/
+func GrantVmGroupAccess(name string) error {
+	// Stat (to determine if `name` is a directory).
+	s, err := os.Stat(name)
+	if err != nil {
+		return errors.Wrapf(err, "%s os.Stat %s", gvmga, name)
+	}
+
+	// Get a handle to the file/directory. Must defer Close on success.
+	fd, err := createFile(name, s.IsDir())
+	if err != nil {
+		return err // Already wrapped
+	}
+	defer syscall.CloseHandle(fd)
+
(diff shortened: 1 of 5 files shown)


pkg/security/grantvmgroupaccess.go
```diff
@@ -0,0 +1,285 @@
...
+func GrantVmGroupAccess(name string) error {
+
+	// Stat (we use this to determine if file/directory is a directory).
+	s, err := os.Stat(name)
+	if err != nil {
+		return errors.Wrapf(err, "%s os.Stat %s", gvmga, name)
+	}
+
+	// Get a handle to the file/directory.
+	fd, err := createFile(name, s.IsDir())
```
jterry75: Do we need dir support? Why not just the file?

pkg/security/grantvmgroupaccess.go
```diff
@@ -0,0 +1,285 @@
...
+	trusteeTypeComputer       trusteeType = 8
+)
+
+// GrantVMGroupAccess sets the DACL for the group SIDs to the specified
+// file or directory. This is a golang re-implementation of the same function
+// in vmcompute, just not exported in RS5. Which kind of sucks. Sucks a lot :/
+func GrantVmGroupAccess(name string) error {
+
+	// Stat (we use this to determine if file/directory is a directory).
+	s, err := os.Stat(name)
```
jterry75: Should we verify `name` ends with `vhd` or `vhdx`?

pkg/security/grantvmgroupaccess.go
jhowardmsft: Doesn't harm. And the HCS implementation copes with both.

pkg/security/grantvmgroupaccess.go
jhowardmsft: Why make this library function less generic? I don't think we should.

pkg/security/syscall_windows.go
```diff
@@ -0,0 +1,9 @@
+package security
+
+//go:generate go run mksyscall_windows.go -output zsyscall_windows.go syscall_windows.go
+
+//sys GetSecurityInfo(handle syscall.Handle, objectType uint32, si uint32, ppsidOwner **uintptr, ppsidGroup **uintptr, ppDacl *uintptr, ppSacl *uintptr, ppSecurityDescriptor *uintptr) (err error) [failretval!=0] = advapi32.GetSecurityInfo
```
jterry75: Can we not export these please

pkg/security/grantvmgroupaccess.go
```diff
@@ -0,0 +1,285 @@
...
+	if err := GetSecurityInfo(fd, uint32(ot), uint32(si), nil, nil, &origDACL, nil, nil); err != nil {
+		return errors.Wrapf(err, "%s GetSecurityInfo %s", gvmga, name)
+	}
+
+	// I don't understand why if retrieving the DACL, there is need to call
+	// LocalFree, but there is if retrieving the SD (which contains the DACL
+	// as part of it). Although the HCS implementation of GrantVmGroupAccess
+	// manipulates the SD, it's MUCH easier in go to manipulate the DACL
+	// to configure the relevant ACEs.
+	// defer syscall.LocalFree((syscall.Handle)(unsafe.Pointer(origDACL)))
```
jterry75: Dont you just want to pass `origDACL`. Not a pointer to it? It is already holding the value of the `HLOCAL`

pkg/security/grantvmgroupaccess_test.go
```diff
@@ -0,0 +1,106 @@
...
+// S-1-15-3-1024-2268835264-3721307629-241982045-173645152-1490879176-104643441-2915960892-1612460704:(I)(R,W)
+// S-1-5-83-1-3166535780-1122986932-343720105-43916321:(I)(R,W)
+
+func TestGrantVmGroupAccess(t *testing.T) {
+	f, err := ioutil.TempFile("", "gvmgafile")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer f.Close()
+	defer os.Remove(f.Name())
```
jterry75: Oh Windows :).

You need to switch the order here. The `os.Remove` will fail because the `f.Close()` hasn't run yet. Maybe make it explicit by a single function
```go
defer func() {
    f.Close()
    os.Remove(f.Name())
}()
```

pkg/security/grantvmgroupaccess_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if err := GrantVmGroupAccess(f.Name()); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := GrantVmGroupAccess(d); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := GrantVmGroupAccess(find.Name()); err != nil {
```
jterry75: Shouldn't `find.txt` get this for free. IE don't call this on `find.Name()` but still verify that it is set?

pkg/security/syscall_windows.go
jhowardmsft: Fixed (push coming soon)

pkg/security/grantvmgroupaccess_test.go
jhowardmsft: Fixed. Push coming soon...

pkg/security/grantvmgroupaccess_test.go
jhowardmsft: Yes, I'm an idiot and got carried away with cut and paste. Fixed. Push coming soon.

jhowardmsft:
> Some feedback

@jterry75 - All feedback addressed 😁

pkg/security/grantvmgroupaccess.go
jhowardmsft: Confirmed by @jstarks that this isn't needed and won't leak. Have just removed this comment.

pkg/security/grantvmgroupaccess.go
jterry75: Because we need it for exactly 1 thing. To set the acl on vhd files. When do we need a folder and all children?

jhowardmsft: You asked me to implement GrantVmGroupAccess in go, which is exactly what I did :smile:. How we use it is separate from what the C implementation does.... Who knows whether directory support might be needed in the future?

jhowardmsft:
@jterry75 I'm merging this. Can you verify the end-to-end through the CD snapshotter? I'll work on the docker LCOW graphdriver.

PR #120, opened 2019-03-22
Move smart ETW field logging to base etw library

kevpar:
This is so we can use the smart logging functionality from hcsshim.


jhowardmsft:
I'm not quite sure of the masterplan here but LGTM

PR #121, opened 2019-03-26
add DialPipeContext

sime1:
This adds the DialPipeContext function, providing a way to expire or cancel pending pipe connections using a context.Context.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -138,22 +138,22 @@ func (s pipeAddress) String() string {
 	return string(s)
 }
 
+//helper function used to try to open the pipe multiple times.
 func tryDialPipe(ctx context.Context, path *string) (syscall.Handle, error) {
 	for {
 		select {
 		case <-ctx.Done():
-			err := ctx.Err()
-			if err == context.DeadlineExceeded {
-				err = ErrTimeout
-			} else {
-				err = &os.PathError{Op: "open", Path: *path, Err: err}
-			}
-			return syscall.Handle(0), err
+			return syscall.Handle(0), ctx.Err()
 		default:
 			h, err := createFile(*path, syscall.GENERIC_READ|syscall.GENERIC_WRITE, 0, nil, syscall.OPEN_EXISTING, syscall.FILE_FLAG_OVERLAPPED|cSECURITY_SQOS_PRESENT|cSECURITY_ANONYMOUS, 0)
-			if err != cERROR_PIPE_BUSY {
-				return h, &os.PathError{Op: "open", Path: *path, Err: err}
+			if err == nil {
+				return h, nil
 			}
+			if err != cERROR_PIPE_BUSY {
+				return h, newOpenError(path, err)
+			}
+			// Wait 10 msec and try again. This is a rather simplistic
+			// view, as we always try each 10 milliseconds.
 			time.Sleep(time.Millisecond * 10)
 		}
 	}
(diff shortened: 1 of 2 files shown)


pipe.go
```diff
@@ -137,33 +138,60 @@ func (s pipeAddress) String() string {
...
+		select {
+		case <-ctx.Done():
+			return syscall.Handle(0), ctx.Err()
+		default:
+			h, err := createFile(*path, syscall.GENERIC_READ|syscall.GENERIC_WRITE, 0, nil, syscall.OPEN_EXISTING, syscall.FILE_FLAG_OVERLAPPED|cSECURITY_SQOS_PRESENT|cSECURITY_ANONYMOUS, 0)
+			if err == nil {
+				return h, nil
+			}
+			if err != cERROR_PIPE_BUSY {
+				return h, newOpenError(path, err)
```
jterry75: Do we need `newOpenError` can we just inline this like it was before?

pipe.go
```diff
@@ -137,33 +138,60 @@ func (s pipeAddress) String() string {
 	return string(s)
 }
 
+//helper function used to try to open the pipe multiple times.
+func tryDialPipe(ctx context.Context, path *string) (syscall.Handle, error) {
```
jterry75: NIT: `// tryDialPipe attempts to dial the pipe at `path` until `ctx` cancellation or timeout.

pipe.go
```diff
@@ -137,33 +138,60 @@ func (s pipeAddress) String() string {
...
-		// view, as we always try each 10 milliseconds.
-		time.Sleep(time.Millisecond * 10)
+func newOpenError(path *string, err error) error {
+	if err != nil {
+		return &os.PathError{Err: err, Op: "open", Path: *path}
 	}
+	return nil
+}
+
+//DialPipeContext connects to a named pipe. ctx can be used to cancel or
```
jterry75: NIT: formatting
```
// DialPipeContext attempts to connect to a named pipe by `path` until `ctx`
// cancellation or timeout.
```

jterry75: I'm wondering if we should return `ErrTimeout` from this as well instead of `context.Canceled`. But since we make sure to change the error for the existing paths I'm not sure it matters that much.

pipe_test.go
```diff
@@ -37,6 +38,26 @@ func TestDialListenerTimesOut(t *testing.T) {
 	}
 }
 
+func TestDialListenerGetsCancelled(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	l, err := ListenPipe(testPipeName, nil)
```
jterry75: Thank you for the test!

Can you also add a timeout test for the `DialPipeContext` call? Similar to `TestDialListenerTimesOut`

sime1:
I inlined the error and fixed the comments. Regarding the test, I'm not quite sure I understood the request, I wrote a test with a `Context` with a timeout, is that what you were asking for?

As for the error returned from `DialPipeContext`, since we pass a `Context` I thought using the standard `Context` errors would be better. IMO we should either use `context.DeadlineExceeded` and `context.Canceled` or `ErrTimeout` and define another error, e.g. `ErrCanceled`; it may be useful being able to know if the context was canceled or timed out.

jterry75:
> I inlined the error and fixed the comments. Regarding the test, I'm not quite sure I understood the request, I wrote a test with a `Context` with a timeout, is that what you were asking for?
> 
> As for the error returned from `DialPipeContext`, since we pass a `Context` I thought using the standard `Context` errors would be better. IMO we should either use `context.DeadlineExceeded` and `context.Canceled` or `ErrTimeout` and define another error, e.g. `ErrCanceled`; it may be useful being able to know if the context was canceled or timed out.

I agree. I like that the `Context` version returns standard `Context` errors as expected.

jterry75:
@jhowardmsft - PTAL this looks good

PR #125, opened 2019-04-08
pipe: Fix ListenPipe/DialPipe race

jstarks:
ListenPipe can fail if there is a concurrent DialPipe because there is a
race window where DialPipe can connect to the initial server named pipe
before it is connected to and closed by ListenPipe.

To fix this, use the lower-level NT API for creating the server pipe,
since this API allows for specifying that a server pipe should initially
be in the disconnected state instead of the listening state. This allows
us to avoid the race condition by creating the pipe in the correct state
initially, so there is no longer a need to create a dummy client
connection.


jhowardmsft:
Looks scarily good to me.

jhowardmsft:
@jterry75 I'll vendor this into hcsshim.

PR #127, opened 2019-04-10
file: make Read() use and update file position

### LINKED ISSUE #126: (*win32File).Read() does not use file position
The `Offset`/`OffsetHigh` fields of `syscall.Overlapped` are never set, so the same bytes are returned from `(*win32File).Read()` over and over again.

tmm1:
fixes #126

### DIFF
--- a/file.go
+++ b/file.go
@@ -219,6 +219,13 @@ func (f *win32File) Read(b []byte) (int, error) {
 		return 0, ErrTimeout
 	}
 
+	curpos, err := syscall.Seek(f.handle, 0, 1)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


tmm1:
Looks like this causes a regression because `win32File` is embedded inside `win32Pipe`, so it has no way to know if its underlying handle is an actual file or not.

```
--- FAIL: TestReadTimeout (0.01s)
    pipe_test.go:136: expected ErrTimeout, got illegal seek
```

jstarks:
Perhaps newWin32File should query the file type (`GetFileType`) and behave accordingly.

PR #128, opened 2019-04-17
winio: Add AF_HYPERV listener

jstarks:


hvsock.go
```diff
@@ -0,0 +1,302 @@
...
+
+	"github.com/Microsoft/go-winio/pkg/guid"
+)
+
+//sys bind(s syscall.Handle, name unsafe.Pointer, namelen int32) (err error) [failretval==socketError] = ws2_32.bind
+
+const (
+	afHvSock = 34 // AF_HYPERV
+
+	socketError = ^uintptr(0)
```
jterry75: What is `^` this syntax?

kevpar: I think that's boolean NOT.

hvsock.go
```diff
@@ -0,0 +1,302 @@
...
+			sock.Close()
+		}
+	}()
+	c, err := l.sock.prepareIo()
+	if err != nil {
+		return nil, l.opErr("accept", err)
+	}
+	defer l.sock.wg.Done()
+
+	const addrlen = 16 + 36
```
jterry75: no `sizeof()`....

hvsock.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	return n, err
+}
+
+func (conn *HvsockConn) Write(b []byte) (int, error) {
+	t := 0
+	for len(b) != 0 {
+		n, err := conn.write(b)
+		if err != nil {
+			return t + n, err
```
jterry75: The protocol allows for a `n` bytes written and an error? I would of assumed (I guess incorrectly) that the behavior here is unknown so a full resend would be expected.

hvsock.go
```diff
@@ -0,0 +1,302 @@
...
+	err := syscall.Shutdown(conn.sock.handle, syscall.SHUT_RD)
+	if err != nil {
+		return os.NewSyscallError("shutdown", err)
+	}
+	return nil
+}
+
+// CloseRead shuts down the read end of the socket.
+func (conn *HvsockConn) CloseRead() error {
+	return conn.opErr("shutdown", conn.shutdown(syscall.SHUT_RD))
```
jterry75: This is a very odd pattern to have the receiver decide on the error if the func fails. Why not:

```go
err := conn.shutdown(syscall.SHUT_RD)
if err != nil {
    return opErr("shutdown", err)
}
return nil
```

Is this just for a single line calling convention?

hvsock.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// RemoteAddr returns the remote address of the connection.
+func (conn *HvsockConn) RemoteAddr() net.Addr {
+	return &conn.remote
+}
+
+// SetDeadline implements the net.Conn SetDeadline method.
+func (conn *HvsockConn) SetDeadline(t time.Time) error {
+	conn.SetReadDeadline(t)
```
jterry75: Seems odd to drop the error here but if I call `SetReadDeadline` or `SetWriteDeadline` by themselves I would get the error.

hvsock.go
jstarks: Binary NOT

jstarks: (I.e. -1)

hvsock.go
jstarks: > go doc io.Writer
type Writer interface {
        Write(p []byte) (n int, err error)
}
    Writer is the interface that wraps the basic Write method.

    Write writes len(p) bytes from p to the underlying data stream. It returns
    the number of bytes written from p (0 <= n <= len(p)) and any error
    encountered that caused the write to stop early. Write must return a non-nil
    error if it returns n < len(p). Write must not modify the slice data, even
    temporarily.

    Implementations must not retain p.

hvsock.go
jstarks: Yeah. I guess this isn't idiomatic. Updated.

hvsock.go
kevpar: Woops, I meant binary

hvsock.go
jterry75: Interesting. I think I got this wrong in another spot. Will go look

hvsock.go
jterry75: Oh lol...

hvsock.go
jstarks: These functions can't actually fail.

PR #130, opened 2019-04-21
Fix ETW package to work on 386

kevpar:
The ETW registration handle is defined to be 64-bits on all platforms. The Go
type previously used a uintptr which created problems on 32-bit systems. This
change adds a new wrapper file which receives a 64-bit handle for the functions
defined in it, and correctly passes it to the native ETW functions either as-is
(on 64-bit) or as two 32-bit values (on 32-bit).

This required a minor change in mksyscall_windows.go, to allow multiple syscalls
that map to the same underlying function without causing a duplicate definition
error in the generated file.

### DIFF
--- a/pkg/etw/etw.go
+++ b/pkg/etw/etw.go
@@ -7,9 +7,14 @@
 // set of C macros.
 package etw
 
-//go:generate go run $GOROOT/src/syscall/mksyscall_windows.go -output zsyscall_windows.go etw.go
+//go:generate go run mksyscall_windows.go -output zsyscall_windows.go etw.go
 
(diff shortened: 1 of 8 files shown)


jstarks:
Could you split the import of Go's mksyscall_windows.go file into a separate commit so that it's clear what changes you have made to it?

kevpar:
> Could you split the import of Go's mksyscall_windows.go file into a separate commit so that it's clear what changes you have made to it?

@jstarks Added the stock mksyscall_windows.go as a new commit, so the diff shows up if you just view the latest commit.

jstarks:
LGTM. It would be nice to improve mksyscall_windows later to generate all this automatically based on the size of the input arguments.

PR #133, opened 2019-05-05
pkg/etwlogrus: Ensure stable event field order

jstarks:
This change updates the logrus hook to sort the incoming logrus data
before passing it to the tracelogging code. This is useful because it
ensures that two instances of the same event have the fields in the same
order, which is necessary to get WPA to line the fields up.

### DIFF
--- a/pkg/etwlogrus/hook.go
+++ b/pkg/etwlogrus/hook.go
@@ -1,6 +1,8 @@
 package etwlogrus
 
 import (
+	"sort"
+
 	"github.com/Microsoft/go-winio/pkg/etw"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


pkg/etwlogrus/hook.go
```diff
@@ -62,13 +64,29 @@ func (h *Hook) Fire(e *logrus.Entry) error {
...
 	fields := make([]etw.FieldOpt, 0, len(e.Data)+1)
-
 	fields = append(fields, etw.StringField("Message", e.Message))
-
-	for k, v := range e.Data {
-		fields = append(fields, etw.SmartField(k, v))
+	for _, k := range names {
+		fields = append(fields, etw.SmartField(k, e.Data[k]))
+	}
+	if errv != nil {
```
jterry75: Not sure we care but if this ever had `data[logrus.ErrorKey] = nil` we would drop it here. Do we want "Error" to be written based on its keys presence or based on the value?

jstarks: Hmm. I guess for consistency I should make it based on presence.

PR #136, opened 2019-05-07
Transition to Go modules

kevpar:


jstarks:
Please don't vendor here; we don't build binaries out of this repo so we don't need vendoring.

kevpar:
@jstarks vendoring removed

Issue #135, opened 2019-05-07
Case-conflict between canonical repo address and go imports


liggitt:
The github.com/Microsoft org was just renamed to github.com/microsoft

This means that the canonical github repo address now conflicts with canonical go import packages (for example, `"github.com/Microsoft/go-winio/pkg/guid"`, as referenced from https://github.com/microsoft/go-winio/blob/master/hvsock.go#L12)

It is now very easy for downstream consumers that depend on leaf packages only (which don't import any other upper-case "github.com/Microsoft/..." packages) to accidentally clone and add imports to lower-case `"github.com/microsoft/..."` packages. This will work in isolation, but as soon as their library has to coexist in a build with *any* other component which used the canonical `"github.com/Microsoft/..."` imports, builds will start failing.

(This was possible before, but way harder to do accidentally, since all the links and clone commands from github agreed with the canonical go import packages)

Context:
* https://twitter.com/ehotinger/status/1125792922463424512
* https://twitter.com/liggitt/status/1125843936864612353
* https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276

Please consider reverting the `Microsoft` -> `microsoft` rename.

liggitt:
cc @jeffwilcox

dims:
cc @michmike

liggitt:
also filed in the other Microsoft dependency we use in Kubernetes: https://github.com/microsoft/hcsshim/issues/591

liggitt:
(hoisting from https://twitter.com/liggitt/status/1125879048658403329 so it doesn't get lost)

If the case change is here to stay, some things that would help a lot:

1) guidance and checks to ensure all the go repos under the org agree and continue to use `Microsoft`
2) example `go get` commands in the READMEs of all the go repos to guide consumers
3) add canonical import paths with the correct case (https://golang.org/doc/go1.4#canonicalimports) to prevent accidental use of the wrong case downstream

PR #137, opened 2019-05-07
[org-rename] Update github org name

bketelsen:
This PR is a brute-force org rename to fix go import paths for the recent Microsoft org rename (Microsoft->microsoft)

### DIFF
--- a/backuptar/tar.go
+++ b/backuptar/tar.go
@@ -14,8 +14,8 @@ import (
 	"syscall"
 	"time"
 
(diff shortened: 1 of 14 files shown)


liggitt:
Thanks for holding this out. See suggestions in https://github.com/microsoft/go-winio/issues/135#issuecomment-490266958 for safeguards to prevent mixed case use from entering the ecosystem. 

Recasing the internal imports like this is exceptionally disruptive

PR #139, opened 2019-05-08
etwlogrus: Fix event logging not including fields

kevpar:


jstarks:
Bizarre--I saw that bug and thought I fixed it, but somehow that got lost. Thanks.

PR #141, opened 2019-05-08
pkg/guid: Prefer passing GUID by value

kevpar:

### DIFF
--- a/pkg/guid/guid.go
+++ b/pkg/guid/guid.go
@@ -12,7 +12,7 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-var _ = (json.Marshaler)(&GUID{})
+var _ = (json.Marshaler)(GUID{})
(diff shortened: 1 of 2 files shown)


jterry75:
LGTM but why?

kevpar:
@jterry75 I was discussing with @jstarks about when to use a GUID by value vs by reference, and we concluded it would be simpler to just use it by value everywhere. As another datapoint, Google's UUID package passes around its `[16]byte` array by value everywhere, which is equivalent to what we're doing here.

We will just need to pass it by reference to the native methods which expect a `LPCGUID`, but that shouldn't be a problem.

jterry75:
I'm fine with it, it's more `go-ish` anyways I was just wondering since we are using `windows.GUID` and it's not `[16]byte`

kevpar:
Yeah, it's kind of a compromise between Windows and Go semantics. Windows format, but prefer pass-by-value like Go.

PR #142, opened 2019-05-08
Fix errors introduced with guid-by-value change

kevpar:


kevpar:
I forgot to test the rest of the guid usage in go-winio, so this fixes the problems I introduced.

jterry75:
Do we not have any tests for this that run on checkin?

kevpar:
> Do we not have any tests for this that run on checkin?

@jterry75 It doesn't look like we have any CI at all for go-winio.

PR #146, opened 2019-06-01
Fix interface-wrapped nil returned by MakeOpenFile

kevpar:


kevpar:
Interested in opinions: better to fix this way, or adjust callers to not use the value at all if an error is returned?

kevpar:
@jterry75 PTAL when you get a chance

kevpar:
Woops, accidentally closed the PR

kevpar:
@jterry75 PTAL

PR #157, opened 2019-08-05
Fix regression in DetachVhd

jterry75:
When opening a disk for detach mode no parameters are required. This regression
was introduced when the OpenVirtualDisk method was added. Moving the code back
to the previous openVirtualDisk internal method so we can pass nil parameters.

Signed-off-by: Justin Terry (VM) <juterry@microsoft.com>

### DIFF
--- a/vhd/vhd.go
+++ b/vhd/vhd.go
@@ -117,9 +117,13 @@ func CreateVhdx(path string, maxSizeInGb, blockSizeInMb uint32) error {
 	return nil
 }
 
-// DetachVhd detaches a VHD attached at the given path.
+// DetachVhd detaches a mounted container layer vhd found at `path`.
 func DetachVhd(path string) error {
-	handle, err := OpenVirtualDisk(path, VirtualDiskAccessDetach, OpenVirtualDiskFlagNone)
+	handle, err := OpenVirtualDisk(
+		path,
+		VirtualDiskAccessNone,
+		OpenVirtualDiskFlagCachedIO|OpenVirtualDiskFlagIgnoreRelativeParentLocator)
+
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


jterry75:
@Random-Liu - FYI

Random-Liu:
LGTM.

Both Docker and containerd is using the version with regression.

kevpar:
If there are any doc links explaining that `DetachVirtualDisk` requires the disk to be opened with null `Parameters`, that would be great to link here.

Random-Liu:
@kevpar Completely agree. I read through the doc https://docs.microsoft.com/en-us/windows/win32/api/virtdisk/nf-virtdisk-openvirtualdisk, but found nothing about it.

Random-Liu:
@jterry75 It seems that you have a new fix. :) Can you briefly explain?

jterry75:
@Random-Liu / @kevpar - Sorry thought that I updated the commit comment as well. I couldn't find any good docs here either. But reading the code it appears that there are a few valid cases for `Detach`.

1. Pass `parameters == nil`. This works but required me to duplicate the `OpenVirtualDisk` code which seemed odd.
2. Pass `parameters, Version: 1`. This works but required me to again duplicate the `OpenVirtualDisk` code because it is hard coded to the `Version == 2` parameters.
3. Pass `parameters, Version: 2, accessMask == VirtualDiskAccessNone`.

I went with 3 because it seemed to be the smallest change that fixes the issue for both `.vhd, .vhdx`.

jterry75:
Hold on this wont work.

jterry75:
Turns out that when we use `ActivateLayer` we set `ATTACH_VIRTUAL_DISK_FLAG_PERMANENT_LIFETIME` which means that although this code does "detach" it doesn't actually remove it. We must pass the `VIRTUAL_DISK_ACCESS_DETACH` flag which means we have to go back to using the `nil` `parameters` fix. Will change back.

jterry75:
@kevpar / @Random-Liu - After speaking with some people internally about this it seems the intention of this code was to replace the call to `vmcompute.DeactivateLayer`. With that frame in mind and verifying where in docker/containerd this code is called we should be using the current fix so PTAL.

ACCESS_MASK has been deprecated which is why this flag must be set to `None` for all `Version == 2` opens. The additional `flag` changes are what the current `vmcompute.DeactivateLayer` code uses which is 100% the same now.

Random-Liu:
LGTM

PR #159, opened 2019-09-12
lowercase module name

### LINKED ISSUE #156: go-winio module problem
I can't install go-winio module.
go version 1.12..7 on Windows, without go-winio on disc.
I create new project and execute:
go mod init test
go get github.com/microsoft/go-winio

> go: github.com/microsoft/go-winio@v0.4.13: parsing go.mod: unexpected module path "github.com/Microsoft/go-winio"
> go: error loading module requirements

balboah:
Breaks builds on non-windows,

`parsing go.mod: unexpected module path "github.com/Microsoft/go-winio"`

Fixes #156

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/Microsoft/go-winio
+module github.com/microsoft/go-winio
 
 go 1.12
 


noxer:
The problems occur on case sensitive file systems. Some of the microsoft packages use package names with `Microsoft`, some with `microsoft` which also leads to problems when modules are vendored on one file system and then copied to another one.

TBBle:
I feel like this change would needlessly break users who are using `github.com/Microsoft/go-winio` as an import correctly, instead of trying to go via `github.com/microsoft/go-winio` as seen in #156.

kevpar:
I agree that this would be too disruptive to existing users to change the module name to lower-case. I think a good change to improve the state of things would be to add a canonical import path of `github.com/Microsoft/go-winio` to the files in this repo.

PR #160, opened 2019-09-18
Add new process package

kevpar:
Adding a new process package to encapsulate functionality for working with processes. These new functions will be used as part of the UVM metrics work in hcsshim.


pkg/process/process.go
```diff
@@ -0,0 +1,86 @@
...
+	if err := getProcessMemoryInfo(process, memCounters, uint32(size)); err != nil {
+		return nil, err
+	}
+	return memCounters, nil
+}
+
+// These constants are used with QueryFullProcessImageName's flags.
+const (
+	// ImageNameFormatWin32Path indicates to format the name as a Win32 path.
+	ImageNameFormatWin32Path = iota
```
jterry75: Dont you need the flag name?

kevpar: What do you mean by flag name?

jterry75: Typically if you wanted to do this you would:
```
type ImageNameFormat int32

const (
    ImageNameFormatWin32Path ImageNameFormat = iota
    ImageNameFormatNTPath
)
```

Without the `typename` I didnt even know this would work. What is the value of `ImageNameFormatNTPath` ? Zero? Shouldnt it be 1?

kevpar: I've seen this kind of format fairly frequently when you don't want to define a full new type (e.g. you want this to be a fairly thin abstraction on the underlying syscall), but do want to define some constants. e.g. golang.org/x/sys/windows has this:

```
const (
    // do not reorder
    NetSetupUnknownStatus = iota
    NetSetupUnjoined
    NetSetupWorkgroupName
    NetSetupDomainName
)
```

The `= iota` should still be inherited by subsequent lines, so `ImageNameFormatNTPath` should be 1.

jterry75: Nvm this is fine. I did not know it worked like this

jterry75: The `iota` expression continues until a new type is encountered so it works with the `default` undefined type as well (int)

kevpar: One note, my understanding is an untyped constant defined as a numerical value will actually not take on any type, rather than defaulting to int. So I think we could actually assign this value to a variable of type `int` or `int8` without a cast.

PR #164, opened 2019-09-26
fix GetCurrentProcess api 

maditya:
per https://go-review.googlesource.com/c/sys/+/196798


jterry75:
Same as #163. Closing out for now.

PR #165, opened 2019-09-30
Move to upstream Sddl/SecurityAttribute functions

zx2c4:
These have been moved into x/sys/windows entirely.


jterry75:
@zx2c4 - The change looks good but I am worried about just deleting the exported functions. I think that we should likely go through a deprecation period. @kevpar / @jstarks Thoughts?

jstarks:
We should leave the existing functions but have them call through to the windows package (and only if the semantics are exactly the same).

katiewasnothere:
Is there any update on this?

thaJeztah:
My comment above is now outdated; it was decided not to revert (https://github.com/moby/moby/pull/40021#issuecomment-551439669)

> This is good to go now that golang/go#34610 has been closed and it was decided there will be no revert (golang/go#34610 (comment))


So from that perspective this should be good to go (I'm not a maintainer here though!)

katiewasnothere:
@jstarks do we want these changes? 

We need to update (for example) hcsshim to no longer use out of date functions such as SddlToSecurityDescriptor so we can update the x/sys dependency there. Might be good then to update this here for consistency.

backuptar/tar.go
```diff
@@ -317,32 +319,34 @@ func FileInfoFromHeader(hdr *tar.Header) (name string, size int64, fileInfo *win
 // tar file that was not processed, or io.EOF is there are no more.
 func WriteBackupStreamFromTarFile(w io.Writer, t *tar.Reader, hdr *tar.Header) (*tar.Header, error) {
```
gogitops: warning: cyclomatic complexity 28 of function WriteBackupStreamFromTarFile() is high (> 15) (gocyclo)

kolyshkin:
> We need to update (for example) hcsshim to no longer use out of date functions such as SddlToSecurityDescriptor so we can update the x/sys dependency there. Might be good then to update this here for consistency.

It was a simple fix, unless I'm missing something; please see the second commit in https://github.com/microsoft/hcsshim/pull/828

katiewasnothere:
I've made a new PR https://github.com/microsoft/go-winio/pull/172 to address a few other things.

PR #166, opened 2019-12-03
Expose pipe  access level to caller

paruizfi:
Created a couple of override methods so that caller has more flexibility on how the pipe is created

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -182,13 +182,14 @@ func (s pipeAddress) String() string {
 }
 
 // tryDialPipe attempts to dial the pipe at `path` until `ctx` cancellation or timeout.
-func tryDialPipe(ctx context.Context, path *string) (syscall.Handle, error) {
+func tryDialPipe(ctx context.Context, path *string, access uint32) (syscall.Handle, error) {
 	for {
+
 		select {
 		case <-ctx.Done():
 			return syscall.Handle(0), ctx.Err()
 		default:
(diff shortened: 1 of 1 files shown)


pipe.go
```diff
@@ -223,9 +224,36 @@ func DialPipe(path string, timeout *time.Duration) (net.Conn, error) {
...
+	conn, err := DialPipeContextAccess(ctx, path, syscall.GENERIC_READ|syscall.GENERIC_WRITE)
+	if err == context.DeadlineExceeded {
+		return nil, ErrTimeout
+	}
+	return conn, err
+}
+
+// DialPipeContext attempts to connect to a named pipe by `path` until `ctx`
+// cancellation or timeout.
+func DialPipeWithAccess(path string, timeout *time.Duration, access uint32) (net.Conn, error) {
```
jterry75: We dont need two overloads of this. The non `context` version is legacy. Just add `DialPipeAccess(ctx context.Context, path string, access uint32)`

pipe.go
```diff
@@ -223,9 +224,36 @@ func DialPipe(path string, timeout *time.Duration) (net.Conn, error) {
 // DialPipeContext attempts to connect to a named pipe by `path` until `ctx`
 // cancellation or timeout.
 func DialPipeContext(ctx context.Context, path string) (net.Conn, error) {
+	conn, err := DialPipeContextAccess(ctx, path, syscall.GENERIC_READ|syscall.GENERIC_WRITE)
+	if err == context.DeadlineExceeded {
```
jterry75: Also for context based methods we dont want to convert to `ErrTimeout`. It is correct to return `ctx.Err()` which is `context.DeadlineExceeded` from `tryDialPipe`. We only convert in the `DialPipe` for legacy reasons.

PR #172, opened 2020-06-02
Update to use upstream sddl/SecurityAttribute but retain old exported functions

katiewasnothere:
Signed-off-by: Kathryn Baldauf <kabaldau@microsoft.com>


pipe.go
```diff
@@ -286,24 +288,24 @@ func makeServerPipeHandle(path string, sd []byte, c *PipeConfig, first bool) (sy
...
 	// The security descriptor is only needed for the first pipe.
 	if first {
 		if sd != nil {
-			len := uint32(len(sd))
+			len := sd.Length()
 			sdb := localAlloc(0, len)
-			defer localFree(sdb)
-			copy((*[0xffff]byte)(unsafe.Pointer(sdb))[:], sd)
+			defer windows.LocalFree(windows.Handle(sdb))
+			copy((*[0xffff]byte)(unsafe.Pointer(sdb))[:len], (*[0xffff]byte)(unsafe.Pointer(sd))[:len])
```
kevpar, line 300: This looks like we are copying the security descriptor into a new buffer and then casting that buffer to be a security descriptor as well. Should we just change `objectAttributes` to take a `windows.SECURITY_DESCRIPTOR` instead?

jstarks: We're fairly inconsistent about following the Go requirements around the use of Go pointers in FFI calls--would we face any possible object lifetime issues by doing this?

jstarks:
What does this change actually solve?

backuptar/tar.go
```diff
@@ -317,32 +319,34 @@ func FileInfoFromHeader(hdr *tar.Header) (name string, size int64, fileInfo *win
...
 	if sdraw, ok := hdr.Winheaders[hdrRawSecurityDescriptor]; ok {
-		sd, err = base64.StdEncoding.DecodeString(sdraw)
+		sdbytes, err := base64.StdEncoding.DecodeString(sdraw)
+		sd = (*windows.SECURITY_DESCRIPTOR)(unsafe.Pointer(&sdbytes[0]))
 		if err != nil {
 			return nil, err
 		}
 	}
-	if len(sd) != 0 {
+	sdLen := sd.Length()
```
kevpar, line 339: This seems risky to me. I think we are effectively trusting the security descriptor (from a tar archive) to tell us its own length, and then doing a copy equal to that length. This could result in malicious input causing us to leak data to the file, or crash the process.

kevpar: I think for the `hdrRawSecurityDescriptor` we should stick to the old code that treats it as a `[]byte`. Then we have a reliable length.

For the `hdrSecurityDescriptor` case, we might need to keep the old function around, as `windows.SecurityDescriptorFromString` doesn't return the length of the data it wrote out.

katiewasnothere:
> What does this change actually solve?

This really just moves the repo passed the breaking changes made to x/sys/windows. Not super critical.

backuptar/tar.go
```diff
@@ -317,32 +319,34 @@ func FileInfoFromHeader(hdr *tar.Header) (name string, size int64, fileInfo *win
...
 			Id:   winio.BackupSecurity,
-			Size: int64(len(sd)),
+			Size: int64(sdLen),
 		}
 		err := bw.WriteHeader(&bhdr)
 		if err != nil {
 			return nil, err
 		}
-		_, err = bw.Write(sd)
+		_, err = bw.Write((*[0xffff]byte)(unsafe.Pointer(sd))[:sdLen])
```
kolyshkin, line 349: I think this needs  a comment to explain where `0xffff` comes from

go.sum
```diff
@@ -16,3 +16,5 @@ golang.org/x/sys v0.0.0-20190507160741-ecd444e8653b h1:ag/x1USPSsqHud38I9BAC88qd
 golang.org/x/sys v0.0.0-20190507160741-ecd444e8653b/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20190916202348-b4ddaad3f8a3 h1:7TYNF4UdlohbFwpNH04CoPMp1cHUZgO1Ebq5r2hIjfo=
 golang.org/x/sys v0.0.0-20190916202348-b4ddaad3f8a3/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20200523222454-059865788121 h1:rITEj+UZHYC927n8GT97eC3zrpzXdb/voyeOuVKS46o=
+golang.org/x/sys v0.0.0-20200523222454-059865788121/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
```
kolyshkin, line 20: did you run `go mod tidy`?

thaJeztah:
@katiewasnothere looks like this needs a rebase now that https://github.com/microsoft/go-winio/pull/175 was merged

TBBle:
> @katiewasnothere looks like this needs a rebase now that #175 was merged

A quick look suggests the conflicts with #175 are just adjacent-line changes in `WriteBackupStreamFromTarFile` and the imports list.

katiewasnothere:
Updated. @dcantah and/or @kevpar could you take another look here?

vhd/vhd.go
```diff
@@ -13,9 +13,9 @@ import (
...
+//sys createVirtualDisk(virtualStorageType *VirtualStorageType, path string, virtualDiskAccessMask uint32, securityDescriptor *windows.SECURITY_DESCRIPTOR, createVirtualDiskFlags uint32, providerSpecificFlags uint32, parameters *CreateVirtualDiskParameters, overlapped *syscall.Overlapped, handle *syscall.Handle) (err error) [failretval != 0] = virtdisk.CreateVirtualDisk
 //sys openVirtualDisk(virtualStorageType *VirtualStorageType, path string, virtualDiskAccessMask uint32, openVirtualDiskFlags uint32, parameters *OpenVirtualDiskParameters, handle *syscall.Handle) (err error) [failretval != 0] = virtdisk.OpenVirtualDisk
-//sys attachVirtualDisk(handle syscall.Handle, securityDescriptor *uintptr, attachVirtualDiskFlag uint32, providerSpecificFlags uint32, parameters *AttachVirtualDiskParameters, overlapped *syscall.Overlapped) (err error) [failretval != 0] = virtdisk.AttachVirtualDisk
+//sys attachVirtualDisk(handle syscall.Handle, securityDescriptor *windows.SECURITY_DESCRIPTOR, attachVirtualDiskFlag uint32, providerSpecificFlags uint32, parameters *AttachVirtualDiskParameters, overlapped *syscall.Overlapped) (err error) [failretval != 0] = virtdisk.AttachVirtualDisk
```
dcantah, line 18: @TBBle The signature for these is changing in this PR anyways so let's make sure we get in the other /x/sys/windows'ify PR shortly after and we can cut a release.

pipe.go
```diff
@@ -286,29 +276,25 @@ func makeServerPipeHandle(path string, sd []byte, c *PipeConfig, first bool) (sy
...
-			if err := rtlDefaultNpAcl(&dacl).Err(); err != nil {
+			dacl := &windows.ACL{}
+			if err := windows.RtlDefaultNpAcl(&dacl); err != nil {
 				return 0, fmt.Errorf("getting default named pipe ACL: %s", err)
 			}
-			defer localFree(dacl)
-
-			sdb := &securityDescriptor{
-				Revision: 1,
-				Control:  cSE_DACL_PRESENT,
```
dcantah, line 310: We don't want to set `windows.SE_DACL_PRESENT` anymore?

vhd/vhd.go
TBBle: Sounds good. I did a quick review previously, and the conflicts are textual-only. i.e. on [this line I changed](https://github.com/microsoft/go-winio/pull/197/files#diff-499bf99d640b6f9658a21a04868f1fe4d2af9ef12fbdeaad6494c3fc01e1178fR17) the type of `handle` and `overlapped` only.

So assuming this lands first, rebasing #197 should only take a few minutes, mostly spend compile-checking everything.

backuptar/tar.go
```diff
@@ -330,32 +331,34 @@ func FileInfoFromHeader(hdr *tar.Header) (name string, size int64, fileInfo *win
...
 			Id:   winio.BackupSecurity,
-			Size: int64(len(sd)),
+			Size: int64(sdLen),
 		}
 		err := bw.WriteHeader(&bhdr)
 		if err != nil {
 			return nil, err
 		}
-		_, err = bw.Write(sd)
+		_, err = bw.Write((*[(1 << 31) - 1]byte)(unsafe.Pointer(sd))[:sdLen])
```
TBBle, line 361: This idiom fails `-gcflags=all=-d=checkptr` as of Go 1.14. In this case, it's just being moved from elsewhere, so it's not an objection to this change, but a reminder that we need to do a pass over the code-base with `checkptr` (or `race` enabled, which includes it) with Go 1.15 or later, and fix occurrences of this.

The fix itself is pretty simple, I did the same pass for hcsshim in https://github.com/microsoft/hcsshim/pull/926, see `Uint16BufferToSlice`.

thaJeztah: > This idiom fails -gcflags=all=-d=checkptr as of Go 1.14. In this case

Oh! It just occurred to me now that there's no CI running in this repository (other than the `license/cla` check); should we add a basic github-action?

thaJeztah:
@katiewasnothere looks like this needs a rebase because https://github.com/microsoft/go-winio/pull/220 was merged

PR #179, opened 2020-11-10
Add new virtdisk calls/refactor vhd package

dcantah:
This change adds bindings for AttachVirtualDisk and GetVirtualDiskPhysicalPath, some helper functions including
one to create a differencing vhd, exports the create and open parameters, exports the create flags and adds a couple
of missing flags. This will be used mainly to go hand in hand with the new computestorage calls added to hcsshim as all of the
vhd creation and setup needs to be done manually for these calls. microsoft/hcsshim#881.

### DIFF
--- a/vhd/vhd.go
+++ b/vhd/vhd.go
@@ -2,150 +2,323 @@
 
 package vhd
 
-import "syscall"
+import (
+	"fmt"
+	"syscall"
+
+	"github.com/Microsoft/go-winio/pkg/guid"
+	"github.com/pkg/errors"
+	"golang.org/x/sys/windows"
+)
 
 //go:generate go run mksyscall_windows.go -output zvhd.go vhd.go
 
-//sys createVirtualDisk(virtualStorageType *virtualStorageType, path string, virtualDiskAccessMask uint32, securityDescriptor *uintptr, flags uint32, providerSpecificFlags uint32, parameters *createVirtualDiskParameters, o *syscall.Overlapped, handle *syscall.Handle) (err error) [failretval != 0] = VirtDisk.CreateVirtualDisk
-//sys openVirtualDisk(virtualStorageType *virtualStorageType, path string, virtualDiskAccessMask uint32, flags uint32, parameters *openVirtualDiskParameters, handle *syscall.Handle) (err error) [failretval != 0] = VirtDisk.OpenVirtualDisk
-//sys detachVirtualDisk(handle syscall.Handle, flags uint32, providerSpecificFlags uint32) (err error) [failretval != 0] = VirtDisk.DetachVirtualDisk
-
-type virtualStorageType struct {
-	DeviceID uint32
-	VendorID [16]byte
-}
+//sys createVirtualDisk(virtualStorageType *VirtualStorageType, path string, virtualDiskAccessMask uint32, securityDescriptor *uintptr, createVirtualDiskFlags uint32, providerSpecificFlags uint32, parameters *CreateVirtualDiskParameters, overlapped *syscall.Overlapped, handle *syscall.Handle) (err error) [failretval != 0] = virtdisk.CreateVirtualDisk
+//sys openVirtualDisk(virtualStorageType *VirtualStorageType, path string, virtualDiskAccessMask uint32, openVirtualDiskFlags uint32, parameters *OpenVirtualDiskParameters, handle *syscall.Handle) (err error) [failretval != 0] = virtdisk.OpenVirtualDisk
+//sys attachVirtualDisk(handle syscall.Handle, securityDescriptor *uintptr, attachVirtualDiskFlag uint32, providerSpecificFlags uint32, parameters *AttachVirtualDiskParameters, overlapped *syscall.Overlapped) (err error) [failretval != 0] = virtdisk.AttachVirtualDisk
+//sys detachVirtualDisk(handle syscall.Handle, detachVirtualDiskFlags uint32, providerSpecificFlags uint32) (err error) [failretval != 0] = virtdisk.DetachVirtualDisk
+//sys getVirtualDiskPhysicalPath(handle syscall.Handle, diskPathSizeInBytes *uint32, buffer *uint16) (err error) [failretval != 0] = virtdisk.GetVirtualDiskPhysicalPath
 
 type (
-	createVirtualDiskFlag uint32
-	VirtualDiskAccessMask uint32
+	CreateVirtualDiskFlag uint32
 	VirtualDiskFlag       uint32
+	AttachVirtualDiskFlag uint32
+	DetachVirtualDiskFlag uint32
+	VirtualDiskAccessMask uint32
 )
 
-const (
-	// Flags for creating a VHD (not exported)
-	createVirtualDiskFlagNone                        createVirtualDiskFlag = 0
-	createVirtualDiskFlagFullPhysicalAllocation      createVirtualDiskFlag = 1
-	createVirtualDiskFlagPreventWritesToSourceDisk   createVirtualDiskFlag = 2
-	createVirtualDiskFlagDoNotCopyMetadataFromParent createVirtualDiskFlag = 4
+type VirtualStorageType struct {
+	DeviceID uint32
+	VendorID guid.GUID
+}
 
-	// Access Mask for opening a VHD
-	VirtualDiskAccessNone     VirtualDiskAccessMask = 0
-	VirtualDiskAccessAttachRO VirtualDiskAccessMask = 65536
-	VirtualDiskAccessAttachRW VirtualDiskAccessMask = 131072
-	VirtualDiskAccessDetach   VirtualDiskAccessMask = 262144
-	VirtualDiskAccessGetInfo  VirtualDiskAccessMask = 524288
-	VirtualDiskAccessCreate   VirtualDiskAccessMask = 1048576
-	VirtualDiskAccessMetaOps  VirtualDiskAccessMask = 2097152
-	VirtualDiskAccessRead     VirtualDiskAccessMask = 851968
-	VirtualDiskAccessAll      VirtualDiskAccessMask = 4128768
-	VirtualDiskAccessWritable VirtualDiskAccessMask = 3276800
-
-	// Flags for opening a VHD
-	OpenVirtualDiskFlagNone                        VirtualDiskFlag = 0
-	OpenVirtualDiskFlagNoParents                   VirtualDiskFlag = 0x1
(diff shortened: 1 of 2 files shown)


dcantah:
What's the policy on breaking existing callers? It's a pain to make multiple similarly named methods just because a previous call took in a string path instead of a handle, or didn't allow customizing the open/create parameters.

kevpar:
We probably can't break existing callers. Can we leave the old callers and just add new functions where needed?

As an aside, this seems like a lot of changes, and it's not clear (there's no description) what the value/need driving this is. Are all of these changes needed? Can you please add a description/summary of what this is needed for?

dcantah:
@kevpar The old signatures were untouched and anything new was just added as new functions. I wanted an answer to that question mainly beforehand but I'll add a summary!

dcantah:
@kevpar Added as the commit message. Do you have any history on why the API for creating is so restricted here currently? The flags and parameters struct are both not exported. Like it says in the commit message I've kept the current calls the same but added new calls where every parameter and flag can be specified, and it's unclear to me whether @ambarve needed the GetVirtualDiskPhysicalPath function for cimfs but it's included in this. I didn't need it for the scratch space work at least.

vhd/vhd.go
```diff
@@ -2,96 +2,159 @@
...
+	ResiliencyGUID           guid.GUID
+}
+
+type CreateVirtualDiskParameters struct {
+	Version  uint32 // Must always be set to 2
+	Version2 CreateVersion2
+}
+
+type OpenVersion2 struct {
+	GetInfoOnly    bool
```
anmaxvl, line 55: similar comment as above, whether it's safe to change the type here.

vhd/vhd.go
```diff
@@ -2,96 +2,159 @@
...
+	Version2 AttachVersion2
 }
 
-// CreateVhdx will create a simple vhdx file at the given path using default values.
+const (
+	VIRTUAL_STORAGE_TYPE_DEVICE_VHDX = 0x3
+
+	// Access Mask for opening a VHD
+	VirtualDiskAccessNone     VirtualDiskAccessMask = 0x00000000
+	VirtualDiskAccessAttachRO VirtualDiskAccessMask = 0x00010000
```
anmaxvl, line 80: 👍 definitely better than in base 10 notation 😄

anmaxvl (commented):
some questions about the struct changes, but looks ok otherwise

vhd/vhd.go
```diff
@@ -2,96 +2,159 @@
...
-)
+type VirtualStorageType struct {
+	DeviceID uint32
+	VendorID guid.GUID
+}
 
-type createVersion2 struct {
-	UniqueID                 [16]byte // GUID
+type CreateVersion2 struct {
+	UniqueID                 guid.GUID
```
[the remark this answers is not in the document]
dcantah: Was confused on the current form of the binding as the type is a standard windows GUID anyways, so seems like the update should be correct? https://docs.microsoft.com/en-us/windows/win32/api/virtdisk/ns-virtdisk-open_virtual_disk_parameters

https://docs.microsoft.com/en-us/windows/win32/api/guiddef/ns-guiddef-guid

vhd/vhd.go
dcantah: AFAIK nothing should break with this. We've typed similar bindings that return the windows type BOOL as actual bools.

vhd/vhd.go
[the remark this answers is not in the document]
anmaxvl: cool, thanks for clarification.

anmaxvl (approved):
lgtm

vhd/vhd.go
```diff
@@ -102,50 +165,186 @@ func CreateVhdx(path string, maxSizeInGb, blockSizeInMb uint32) error {
...
-		return 0, err
+		&handle,
+	); err != nil {
+		return syscall.Handle(handle), errors.Wrap(err, "failed to open virtual disk")
+	}
+	return syscall.Handle(handle), nil
+}
+
+// OpenVirtualDiskWithParameters obtains a handle to a VHD opened with supplied access mask and flags.
+func OpenVirtualDiskWithParameters(vhdPath string, virtualDiskAccessMask VirtualDiskAccessMask, openVirtualDiskFlags VirtualDiskFlag, parameters *OpenVirtualDiskParameters) (syscall.Handle, error) {
```
ambarve, line 267: Was this added because you need to open a vhd by specifying different parameters than the default ones?

vhd/vhd.go
```diff
@@ -102,50 +165,186 @@ func CreateVhdx(path string, maxSizeInGb, blockSizeInMb uint32) error {
...
+		parameters,
+		&handle,
+	); err != nil {
+		return handle, errors.Wrap(err, "failed to open virtual disk")
+	}
+	return handle, nil
+}
+
+// CreateVirtualDisk creates a virtual harddisk and returns a handle to the disk.
+func CreateVirtualDisk(path string, virtualDiskAccessMask VirtualDiskAccessMask, createVirtualDiskFlags CreateVirtualDiskFlag, parameters *CreateVirtualDiskParameters) (syscall.Handle, error) {
```
ambarve, line 286: Same question as `OpenVirtualDiskWithParameters` for this.

vhd/vhd.go
```diff
@@ -102,50 +165,186 @@ func CreateVhdx(path string, maxSizeInGb, blockSizeInMb uint32) error {
...
+		parameters,
+		nil,
+		&handle,
+	); err != nil {
+		return handle, errors.Wrap(err, "failed to create virtual disk")
 	}
 	return handle, nil
 }
+
+// GetVirtualDiskPhysicalPath takes a handle to a virtual hard disk and returns the physical
```
ambarve, line 308: Since this is a public API it would be worth it to add an example of what a physical path looks like.

dcantah: Good call, adding

vhd/vhd.go
dcantah: We need this so we can pass `ParentPath` in `CreateVirtualDiskParameters` so we can create the differencing disks for the base layers.

vhd/vhd.go
dcantah: No use-case for it for any of my work but we're already adding the option for the create side so I'd rather just add it to avoid a future change where it may be needed

ambarve:
LGTM!

vhd/vhd.go
```diff
@@ -2,96 +2,159 @@
...
+	AttachVirtualDiskFlagNonPnp                        AttachVirtualDiskFlag = 0x00000040
+	AttachVirtualDiskFlagRestrictedRange               AttachVirtualDiskFlag = 0x00000080
+	AttachVirtualDiskFlagSinglePartition               AttachVirtualDiskFlag = 0x00000100
+	AttachVirtualDiskFlagRegisterVolume                AttachVirtualDiskFlag = 0x00000200
+
+	// Flags for detaching a VHD
+	DetachVirtualDiskFlagNone DetachVirtualDiskFlag = 0x0
+)
+
+var (
```
katiewasnothere, line 133: Can/should these be consts?

vhd/vhd.go
```diff
@@ -102,50 +165,187 @@ func CreateVhdx(path string, maxSizeInGb, blockSizeInMb uint32) error {
...
 }
 
 // OpenVirtualDisk obtains a handle to a VHD opened with supplied access mask and flags.
-func OpenVirtualDisk(path string, accessMask VirtualDiskAccessMask, flag VirtualDiskFlag) (syscall.Handle, error) {
-	var (
-		defaultType virtualStorageType
-		handle      syscall.Handle
-	)
-	parameters := openVirtualDiskParameters{Version: 2}
+func OpenVirtualDisk(vhdPath string, virtualDiskAccessMask VirtualDiskAccessMask, openVirtualDiskFlags VirtualDiskFlag) (syscall.Handle, error) {
```
katiewasnothere, line 250: Could this function not just call into `OpenVirtualDiskWithParameters` with the constructed `parameters`?

vhd/vhd.go
```diff
@@ -102,50 +165,187 @@ func CreateVhdx(path string, maxSizeInGb, blockSizeInMb uint32) error {
...
+}
+
+// CreateDiffVhd is a helper function to create a differencing virtual disk.
+func CreateDiffVhd(diffVhdPath, baseVhdPath string) error {
+	// Setting `ParentPath` is how to signal to create a differencing disk.
+	createParams := &CreateVirtualDiskParameters{
+		Version: 2,
+		Version2: CreateVersion2{
+			ParentPath:       windows.StringToUTF16Ptr(baseVhdPath),
+			BlockSizeInBytes: 1 * 1024 * 1024,
```
katiewasnothere, line 333: Is there a reason for this specific size?

dcantah: Nope not really. Just what HCS used for the diff VHDs but if we're going down the path of having most things configurable, I'll add a `blockSizeInMb` argument to this like there is for `CreateVhdx`

vhd/vhd.go
dcantah: Yea it definitely could, would you prefer? Just kinda left this alone other than changing to use the new storage type.

katiewasnothere: I mean it calls the same thing so I don't see an issue.

vhd/vhd.go
dcantah: Go doesn't support struct constants

vhd/vhd.go
dcantah: Added an option to configure this

vhd/vhd.go
dcantah: Changed

vhd/vhd.go
```diff
@@ -102,50 +165,185 @@ func CreateVhdx(path string, maxSizeInGb, blockSizeInMb uint32) error {
...
+		OpenVirtualDiskFlagCachedIO|OpenVirtualDiskFlagIgnoreRelativeParentLocator,
+	)
 	if err != nil {
 		return err
 	}
+
 	defer syscall.CloseHandle(handle)
-	return detachVirtualDisk(handle, 0, 0)
+	params := &AttachVirtualDiskParameters{Version: 2}
+	if err := attachVirtualDisk(
```
katiewasnothere, line 236: Can we call into `AttachVirtualDisk` here?

dcantah: Done

katiewasnothere (approved):
LGTM

PR #185, opened 2020-12-12
Implement `winio.GetFileStandardInfo`

TBBle:
This get-only API will be useful for detecting when a file has multiple links to it.

Specifically, I plan to use this to provide an optimisation for https://github.com/microsoft/hcsshim/pull/901 (or afterwards, it's not a blocker...) by only attempting to recognise hard-links for files which have multiple links.

See https://github.com/TBBle/hcsshim/compare/base-layer-manipulation...TBBle:hack-base-layer-manipulation-with-branched-go-winio for the example use-case.


fileinfo.go
```diff
@@ -43,6 +67,46 @@ func SetFileBasicInfo(f *os.File, bi *FileBasicInfo) error {
...
+	}
+	if raw.Directory != 0 {
+		cooked.Directory = true
+	}
+	return cooked
+}
+
+// GetFileStandardInfo retrieves ended information for the file.
+func GetFileStandardInfo(f *os.File) (*FileStandardInfo, error) {
+	si := &rawFileStandardInfo{}
```
anmaxvl, line 102: do we need to use an intermediate struct here? can't this be just `&FileStandardInfo{}`?

TBBle: I did this to get the `bool` in the public API to save lots of `!=0` checks for the users. I'm not sure if a `bool` in Go would match a `uint32 /* BOOLEAN */` safely when we're passing unsafe pointers into the C API.

anmaxvl: the public API makes sense. we did have a recent change, where we wrapped `BOOLEAN` with `bool`: https://github.com/microsoft/go-winio/blob/master/vhd/vhd.go#L54-L58.

anmaxvl: maybe @dcantah has more thoughts on this

dcantah: The OpenVirtualDisk options take in `BOOL`'s which is a typedef for int so I'm probably in the wrong here. For type BOOL for functions we'd been using just bool but for these we'd need the padding.

dcantah (approved):

fileinfo.go
TBBle: Yeah, looking at `OpenVirtualDisk2` and how it flows into `_openVirtualDisk`, I suspect that if `procOpenVirtualDisk` took a "size" parameter for the parameters, it would have been failing because the struct was too small.

TBBle:
The discussion about bools reminded me to that I hadn't written any tests; It turns out the `Directory` and `DeletePending` flags are possibly swapped or otherwise faulty. I'm looking into it now, will repush once I have fixed it, and with tests. Marking this draft in the mean-time.

fileinfo.go
TBBle: So after all that, it turns out `BOOLEAN` _is_ 1 byte, not four. I don't know where I got `int32` from before, but the [protocol spec for driver developers](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/5afa7f66-619c-48f3-955f-68c4ece704ae) shows that it's two bytes for two booleans, followed by two bytes of padding.

dcantah: Possibly got `BOOL` and `BOOLEAN` mixed up as I tend to do regularly... haha. Although `BOOL` isn't unsigned so that parts a mystery

anmaxvl: ok... `BOOL` vs `BOOLEAN`

TBBle:
I was able to add tests for everything but the `DeletePending` field, assuming that a directory never has an `AllocationSize` or `EndOfFile` value other than 0. (Creating files in the directory didn't affect it... I wonder if an alternative data stream would work...)

As far as I can tell, the only way to trigger `DeletePending` is `SetFileInformationByHandle` with `FileDispositionInfo`. I tried `FILE_FLAG_DELETE_ON_CLOSE` and it still didn't appear to be true. (Unless the structure is still wrong somehow)

anmaxvl (approved):

fileinfo.go
```diff
@@ -43,6 +67,25 @@ func SetFileBasicInfo(f *os.File, bi *FileBasicInfo) error {
 	return nil
 }
 
+// FileStandardInfo contains extended information for the file.
+// FILE_STANDARD_INFO in WinBase.h
+// https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/5afa7f66-619c-48f3-955f-68c4ece704ae
+type FileStandardInfo struct {
+	AllocationSize, EndOfFile int64
+	NumberOfLinks             int32
```
kevpar, line 75: Looks like this should be `uint32`.

https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_file_standard_information

fileinfo.go
```diff
@@ -13,8 +13,32 @@ import (
 //sys setFileInformationByHandle(h syscall.Handle, class uint32, buffer *byte, size uint32) (err error) = SetFileInformationByHandle
 
 const (
-	fileBasicInfo = 0
-	fileIDInfo    = 0x12
+	// FILE_INFO_BY_HANDLE_CLASS
```
kevpar, line 16: Rather than redefining these all here, this might be a good opportunity to just move to using the definitions from here: https://pkg.go.dev/golang.org/x/sys/windows#FileStandardInfo

fileinfo.go
TBBle: Ah, so it is. It even says that in the link directly above the structure, I think I flipped it by accident when I flipped the previous two, which I originally recall writing as `uint64`.

fileinfo.go
TBBle: Oh! I had no idea those were there. Yeah, _much_ better.

TBBle: This turned into an adventure, because those constants needed a newer version of `golang.org/x/sys/windows` than we had. And that also gave us implementations of the syscalls themselves, rather than having to generate them here.

And then I got excited.

TBBle:
Flicking over to take advantage of "golang.org/x/sys/windows" ended up in a bit of a rabbit hole.

I wasn't sure if it's actually desirable to move from (Windows-specific) `syscall` to `golang.org/x/sys/windows`, so I did it just for `go generate` (because the mkwinsyscall script in Go 1.15 made me) and fileinfo.go.

I did _also_ convert the entire rest of the package, and if this is something that we actually want, rather than just a learning exercise for me, the branch is https://github.com/TBBle/go-winio/compare/GetFileStandardInfo...TBBle:use-x_sys_windows-instead-of-syscall, and I can make it a PR after this lands. The top commit still needs work, as `pkg/fs/GetFileSystemType` handles the return value from `GetVolumeInformation` differently from [the documentation](https://docs.microsoft.com/en-au/windows/win32/api/fileapi/nf-fileapi-getvolumeinformationw#return-value) and I want to write a test to validate this. But the rest of the conversion would be ready now. I've only put it through the unit tests locally, but I did manually inspect every substitution except sometimes when I was replacing one generated function with another, and they had the same parameters.

On the other hand, if this PR just got too complex, it's easy enough to roll back to the last version and apply the `uint32` fix that was probably the only blocker for this PR earlier.

ambarve (approved):
LGTM. Just a minor nit.

fileinfo.go
```diff
@@ -36,13 +29,32 @@ func GetFileBasicInfo(f *os.File) (*FileBasicInfo, error) {
...
+	if err := windows.SetFileInformationByHandle(windows.Handle(f.Fd()), windows.FileBasicInfo, (*byte)(unsafe.Pointer(bi)), uint32(unsafe.Sizeof(*bi))); err != nil {
 		return &os.PathError{Op: "SetFileInformationByHandle", Path: f.Name(), Err: err}
 	}
 	runtime.KeepAlive(f)
 	return nil
 }
 
+// FileStandardInfo contains extended information for the file.
+// FILE_STANDARD_INFO in WinBase.h
+// https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/5afa7f66-619c-48f3-955f-68c4ece704ae
```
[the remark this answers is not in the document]
TBBle: Sure. Will do that tonight. I liked the current link because I don't actually know off-hand the size/signedness of `LARGE_INTEGER`, `DWORD`, and `BOOLEAN`, and that link gives me bit-counts and signedness.

TBBle:
Poke: I think this one's ready to go, and I'd like to land it so I can update https://github.com/microsoft/hcsshim/pull/901 to take advantage of it, as well as flush it through so I can rebase my "Kill `syscall`" branch as a PR for consideration.

thaJeztah:
@kevpar ptal 🤗

katiewasnothere (approved):
LGTM, could you rebase?

dcantah (approved):

TBBle:
Rebased. Helpfully someone pulled in an even-newer `golang.org/x/sys` so the changes got shorter. ^_^

PR #188, opened 2021-01-16
Correct GetVolumeInformation return type handling

TBBle:
The [docs](https://msdn.microsoft.com/en-us/library/windows/desktop/aa364993(v=vs.85).aspx) say it returns 0 for failure, non-zero for success.

This implementation is looking for <0 as HRESULT failure, and deriving
an Errno from that.

The new tests fail with GetFileSystemType as it was defined, e.g.,
```
--- FAIL: TestGetFSTypeOfValidButAbsentDrive (0.00s)
    fs_windows_test.go:41: GetFileSystemType a:\ unexpectedly succeeded
```
when I definitely do not have an A:\ drive.

### DIFF
--- a/pkg/fs/fs_windows.go
+++ b/pkg/fs/fs_windows.go
@@ -3,8 +3,6 @@ package fs
 import (
 	"errors"
 	"path/filepath"
-	"syscall"
-	"unsafe"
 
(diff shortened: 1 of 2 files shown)


katiewasnothere (approved):
LGTM

PR #192, opened 2021-01-27
Add ability to set provider group

kylewo:
Adding the ability for a caller of the ETW library to specify a provider group for their ETW provider.

### DIFF
--- a/pkg/etw/provider.go
+++ b/pkg/etw/provider.go
@@ -123,12 +123,6 @@ func NewProvider(name string, callback EnableCallback) (provider *Provider, err 
 	return NewProviderWithOptions(name, WithCallback(callback))
 }
 
-// NewProviderWithGroup creates and registers a new ETW provider with a
-// specificed group. The provider ID is generated based on the provider name.
-func NewProviderWithGroup(name string, group guid.GUID, callback EnableCallback) (provider *Provider, err error) {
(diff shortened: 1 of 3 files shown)


pkg/etw/provider.go
```diff
@@ -120,7 +120,13 @@ func providerIDFromName(name string) guid.GUID {
 // NewProvider creates and registers a new ETW provider. The provider ID is
 // generated based on the provider name.
 func NewProvider(name string, callback EnableCallback) (provider *Provider, err error) {
-	return NewProviderWithID(name, providerIDFromName(name), callback)
+	return NewProviderWithOptions(name, WithCallback(callback))
+}
+
+// NewProviderWithGroup creates and registers a new ETW provider with a
+// specificed group. The provider ID is generated based on the provider name.
+func NewProviderWithGroup(name string, group guid.GUID, callback EnableCallback) (provider *Provider, err error) {
```
kevpar, line 128: I think it's fine to not add this. We can consider `NewProvider` and `NewProviderWithID` as legacy, and treat `NewProviderWithOptions` as the single entry-point for creating a provider going forward.

kevpar (commented):
Everything looks good, just had one comment about removing a function that I don't think we need.

kevpar (approved):
LGTM. Thanks for contributing!

PR #200, opened 2021-03-24
pkg/etw: Fix to build on windows/arm

kevpar:
We don't support actually using ETW on windows/arm, but to make things
easier for downstream dependencies, we want to still allow the package
to compile and just no-op on this architecture. We do this by returning
a nil Provider, and implementing its methods to no-op when the receiver
is nil.

Previously this was implemented by putting NewProviderWithID in
provider_unsupported.go, but when we refactored the code so that the
actual work was done in NewProviderWithOptions instead, we didn't fix up
provider_unsupported. This change fixes this by putting only
NewProviderWithOptions in provider_unsupported.go, since the other
provider creation functions call into this one.

### DIFF
--- a/pkg/etw/newprovider.go
+++ b/pkg/etw/newprovider.go
@@ -12,46 +12,6 @@ import (
 	"golang.org/x/sys/windows"
 )
 
-type providerOpts struct {
-	callback EnableCallback
-	id       guid.GUID
-	group    guid.GUID
-}
-
-// ProviderOpt allows the caller to specify provider options to
-// NewProviderWithOptions
-type ProviderOpt func(*providerOpts)
-
-// WithCallback is used to provide a callback option to NewProviderWithOptions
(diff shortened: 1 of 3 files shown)


kevpar:
Tested building the following configurations:

```
/mnt/c/Users/kevpar/Source/repos/go-winio/pkg/etw/sample$ GOOS=windows GOARCH=amd64 go build .
/mnt/c/Users/kevpar/Source/repos/go-winio/pkg/etw/sample$ GOOS=windows GOARCH=arm GOARM=7 go build .
/mnt/c/Users/kevpar/Source/repos/go-winio/pkg/etw/sample$ GOOS=linux GOARCH=arm GOARM=7 go build .
```

cpuguy83 (approved):
LGTM

Tested this downstream with containerd.
Fixes our issue! ❤️

dcantah (approved):
lgtm

PR #201, opened 2021-04-05
Add security privilege (needed for VHDX support for docker data root)

awmirantis:
To support using VHDX for the Docker data root and the writing of image layers an additional privilege is needed.  Added "SeSecurityPrivilege" as an additional option.

Microsoft Case#120082125001494
Signed-off-by: Adam Williams <awilliams@mirantis.com>

### DIFF
--- a/privilege.go
+++ b/privilege.go
@@ -28,8 +28,9 @@ const (
 
 	ERROR_NOT_ALL_ASSIGNED syscall.Errno = 1300
 
-	SeBackupPrivilege  = "SeBackupPrivilege"
-	SeRestorePrivilege = "SeRestorePrivilege"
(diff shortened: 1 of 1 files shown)


dcantah (approved):

awmirantis:
@dcantah It appears there was a release yesterday, any idea on the timeline for a release with this change?

dcantah:
@awmirantis We don't have a set release cadence, we just cut releases if they're needed for a consumer of this (which it sounds like Moby needs this so this fits the bill). I can cut a new release with the change shortly.

dcantah:
@awmirantis https://github.com/microsoft/go-winio/releases/tag/v0.4.18 Here you go!

PR #204, opened 2021-04-21
[Temporary] Revert Implement winio.GetFileStandardInfo FileInfo commits 

katiewasnothere:
Addresses issue mentioned in https://github.com/moby/moby/pull/42307


katiewasnothere:
@TBBle @awmirantis @thaJeztah

thaJeztah:
Thank you 🙏 

Was it a clean revert? (making sure if reviewers need to pay extra attention to the changes)

katiewasnothere:
> Was it a clean revert? (making sure if reviewers need to pay extra attention to the changes)

@thaJeztah yep! just git reverted and that was it.

thaJeztah (approved):
🥳 those are the best reverts; in that case

LGTM (but good to have a peek from @TBBle)

dcantah (approved):
Thanks Kathryn :)

TBBle (approved):
A clean revert is what I expected, since these APIs were introduced to support https://github.com/microsoft/hcsshim/pull/901 and that hasn't landed yet, so the only extant users were the tests, also reverted here.

Maybe title the PR "_Temporarily_ revert ..." so no one gets the impression that we don't want this API after all? ^_^

PR #205, opened 2021-04-21
Add CI github action for testing on push and PR

katiewasnothere:
This PR additionally updates the `TestGrantVmGroupAccess` test to account for running on a machine that doesn't have support for running hyper-v VMs, which is the current case for github hosted windows images.

Signed-off-by: Kathryn Baldauf <kabaldau@microsoft.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,12 @@
+name: CI 
+on:
+  - push
+  - pull_request
+
+jobs: 
+  test:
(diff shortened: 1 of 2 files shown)


dcantah:
Can we add `-gcflags=all=-d=checkptr` to the go test? It's caught a lot of stuff for hcsshim that we fixed

dcantah (approved):
Thanks! LGTM

PR #208, opened 2021-04-22
Build the three binaries in this repo from the ci

dcantah:
Even though this repo is normally used as a dependency, there are some sample binaries and tools built out of here. This just makes sure they still build after changes 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,12 +1,24 @@
-name: CI 
+name: CI
 on:
   - push
   - pull_request
 
-jobs: 
+jobs:
   test:
     runs-on: 'windows-2019'
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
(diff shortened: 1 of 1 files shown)


anmaxvl (approved):

.github/workflows/ci.yml
```diff
@@ -1,12 +1,24 @@
...
       - uses: actions/setup-go@v2
       - run: go test -gcflags=all=-d=checkptr -v ./...
+
+  build:
+    runs-on: 'windows-2019'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15.0'
```
katiewasnothere, line 20: we should update the go version in the modules too, but I'm not sure if we want to do that before we make a 0.4.x release to address the breaking changes in moby https://github.com/moby/moby/pull/42307

PR #210, opened 2021-04-26
Fix syscall error handling in vhd and pkg/security

kevpar:
These packages had incorrect error handling for their generated syscall
bindings. The functions they were calling returned errors directly, but
the binding was written such that the generated code was calling
GetLastError instead.

Thankfully, this did not affect the detection of whether or not an error
had occurred, it only caused the value returned in the case of an error
to be not the right error code.


dcantah (approved):

katiewasnothere (approved):
lgtm

PR #211, opened 2021-05-05
Add etw UnicodeStringField

pavelbazika:
Allow to send UTF-16 encoded message. It's the only missing part needed for writing messages with local characters into event log via this library, when string is specified as win:UnicodeString in instrumentation manifest (which cannot be changed).

`UnicodeStringArray` is there only for completeness from my point of view.

### DIFF
--- a/pkg/etw/eventdata.go
+++ b/pkg/etw/eventdata.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"syscall"
+	"unicode/utf16"
 )
 
 // eventData maintains a buffer which builds up the data for an ETW event. It
@@ -26,6 +27,13 @@ func (ed *eventData) toBytes() []byte {
 func (ed *eventData) writeString(data string) {
 	_, _ = ed.buffer.WriteString(data)
 	_ = ed.buffer.WriteByte(0)
+}
+
+// writeUnicodeString appends a string converted to UTF-16, including the null terminator, to the buffer.
+func (ed *eventData) writeUnicodeString(data string) {
+	unicode := utf16.Encode([]rune(data))
+	binary.Write(&ed.buffer, binary.LittleEndian, unicode)
+	ed.buffer.Write([]byte{0, 0})
 }
 
 // writeInt8 appends a int8 to the buffer.
--- a/pkg/etw/fieldopt.go
+++ b/pkg/etw/fieldopt.go
@@ -64,6 +64,14 @@ func JSONStringField(name string, value string) FieldOpt {
 	}
 }
 
+// UnicodeStringField adds a single UTF-16 string field to the event.
+func UnicodeStringField(name string, value string) FieldOpt {
+	return func(em *eventMetadata, ed *eventData) {
+		em.writeField(name, inTypeUnicodeString, outTypeString, 0)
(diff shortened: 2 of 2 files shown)


anmaxvl (approved):
LGTM. can you commit with `-s`?

dcantah (approved):
lgtm also, same ask as above though. Can do `git commit --amend --sign-off` and repush.

pkg/etw/fieldopt.go
```diff
@@ -66,6 +74,17 @@ func StringArray(name string, values []string) FieldOpt {
 	}
 }
 
+// UnicodeStringArray adds an array of UTF-16 strings to the event.
+func UnicodeStringArray(name string, values []string) FieldOpt {
+	return func(em *eventMetadata, ed *eventData) {
+		em.writeArray(name, inTypeUnicodeString, outTypeString, 0)
```
kevpar, line 80: Do we want the out type here? It looks like `TraceLoggingWideString` in `TraceLoggingProvider.h` just uses `TlgInUNICODESTRING` with no out type.

pkg/etw/fieldopt.go
```diff
@@ -55,6 +55,14 @@ func StringField(name string, value string) FieldOpt {
 	}
 }
 
+// UnicodeStringField adds a single UTF-16 string field to the event.
+func UnicodeStringField(name string, value string) FieldOpt {
+	return func(em *eventMetadata, ed *eventData) {
+		em.writeField(name, inTypeUnicodeString, outTypeString, 0)
```
kevpar, line 61: Same question here about the out type.

pavelbazika: Sorry for long delay, I missed the notification.

I followed the manifest here, which states
`<data name="Message" inType="win:UnicodeString" outType="xs:string"/>`

Unfortunately I don't have enough ETW technical background to know if `outTypeDefault` would suffice. The most I can do is to test if it works with `outTypeDefault`, but I can't be sure about some corner cases.

pkg/etw/fieldopt.go
pavelbazika: Please see below.

helsaawy:
@pavelbazika
If you are still interested in merging this, could you rebase and push? it should be pretty quick to merge after that

pkg/etw/eventdata.go
```diff
@@ -28,6 +29,13 @@ func (ed *eventData) writeString(data string) {
 	_ = ed.buffer.WriteByte(0)
 }
 
+// writeUnicodeString appends a string converted to UTF-16, including the null terminator, to the buffer.
+func (ed *eventData) writeUnicodeString(data string) {
+	unicode := utf16.Encode([]rune(data))
+	binary.Write(&ed.buffer, binary.LittleEndian, unicode)
+	ed.buffer.Write([]byte{0, 0})
```
helsaawy, line 36: to suppress the linter on error checking (or you can add a `//nolint:errcheck` directive)
```suggestion
	_ = binary.Write(&ed.buffer, binary.LittleEndian, unicode)
	_, _ = ed.buffer.Write([]byte{0, 0})
```

kevpar:
Closing old PR. Please reactivate if this is still relevant.

PR #213, opened 2021-05-26
Cleanup wim package for use on non Windows platforms

kenshaw:
Cleans up the wim package (removing unnecessary build tag guards), and
cleans up the wim/validate command to be slightly more "Go idiomatic".

### DIFF
--- a/wim/decompress.go
+++ b/wim/decompress.go
@@ -1,5 +1,3 @@
-// +build windows
-
 package wim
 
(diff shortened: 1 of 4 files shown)


kenshaw:
Submitted this as a separate PR in hopes if the package renaming is not accepted, that this would be. Both PRs should be rebase-able against the `master` branch.

PR #214, opened 2021-06-14
Fix corruption in x86 callback

kylewo:
When running on x86, the provider callback doesn't function correctly because two of the arguments are ULONGLONG's (64-bit), which does not align with the uintptr on x86 (32-bit).

This change splits the function into two platform-specific versions to avoid the issue.


kevpar (approved):
LGTM. Nice find!

dcantah (approved):

PR #217, opened 2021-08-16
Fixing corruption in callbacks introduced by x86 changes

kylewo:
In [my previous PR](https://github.com/microsoft/go-winio/pull/214), when breaking out providerCallbackAdapter into separate 32 and 64 bit implementations, I changed the typing of the state and level arguments erroneously. These parameters should be treated as pointers, not as raw integers. This issue will cause the parameters passed into providerCallback to be corrupted.

Typically this is benign, however it causes a crash in the following scenario:
- Callbacks are enabled (i.e. someone is listening for the event provider)
- An event provider was previous registered and unregistered in the same executable.

Under these conditions, the global provider map (in providerglobal.go) will not have an entry at index 0 (the index only ever increases, so the previous registering/unregistering of a provider leaves the 0 slot empty). The corrupted arguments into providerCallback will have an index of 0 instead of the correct index, and then providerCallback (provider.go line 69) will get a null provider and then attempt to update its state, causing a null reference panic.

This was missed in my initial testing, because for the test providers nothing is listening for the events, so the callback never gets called.

### DIFF
--- a/pkg/etw/wrapper_64.go
+++ b/pkg/etw/wrapper_64.go
@@ -46,7 +46,7 @@ func eventSetInformation(
 // for provider notifications. Because Go has trouble with callback arguments of
 // different size, it has only pointer-sized arguments, which are then cast to
 // the appropriate types when calling providerCallback.
-func providerCallbackAdapter(sourceID *guid.GUID, state uint32, level uint8, matchAnyKeyword uintptr, matchAllKeyword uintptr, filterData uintptr, i uintptr) uintptr {
+func providerCallbackAdapter(sourceID *guid.GUID, state uintptr, level uintptr, matchAnyKeyword uintptr, matchAllKeyword uintptr, filterData uintptr, i uintptr) uintptr {
 	providerCallback(*sourceID, ProviderState(state), Level(level), uint64(matchAnyKeyword), uint64(matchAllKeyword), filterData, i)
 	return 0
 }
--- a/pkg/etw/sample/sample.go
+++ b/pkg/etw/sample/sample.go
@@ -7,6 +7,7 @@ import (
 	"bufio"
 	"fmt"
 	"os"
+	"runtime"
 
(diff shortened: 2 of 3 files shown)


pkg/etw/wrapper_64.go
```diff
@@ -46,7 +46,7 @@ func eventSetInformation(
 // for provider notifications. Because Go has trouble with callback arguments of
 // different size, it has only pointer-sized arguments, which are then cast to
 // the appropriate types when calling providerCallback.
-func providerCallbackAdapter(sourceID *guid.GUID, state uint32, level uint8, matchAnyKeyword uintptr, matchAllKeyword uintptr, filterData uintptr, i uintptr) uintptr {
```
kevpar, line 49: I'm a bit confused as to why we need to change these types. From [PENABLECALLBACK](https://docs.microsoft.com/en-us/windows/win32/api/evntprov/nc-evntprov-penablecallback) it looks like they are `ULONG` and `UCHAR`, which should correspond to `uint32` and `uint8`.

kylewo: So this is actually reverting the types to what they were before my initial changes. The comment above explains why they were that way. Go appears to give us all the callback variables as pointers, so if you leave level as a uint8, then part of the level gets shoved into matchAnyKeyword and everything becomes misaligned, causing the corruption.

kevpar: Ah, I think I see now, this is presumably something to do with how Go's callback code is translating the native args to the Go ABI?

kevpar (approved):
LGTM

kevpar:
Oh, one thing. Can you please certify your commit with the following command?

```
git commit --amend -s --no-edit
git push --force
```

dcantah (approved):
lgtm too, barring the same sign-off/certify comment from Kevin. Thanks!

PR #220, opened 2021-09-12
functions SDDL, EA & Reparse encoding

ambarve:
Separates out Security Descriptor (SDDL), Extended Attribute (EA) and Reparse tag encoding
code into individual functions so that those functions can be reused in other code (like
CimFS layer writing)

Signed-off-by: Amit Barve <ambarve@microsoft.com>

### DIFF
--- a/backuptar/tar.go
+++ b/backuptar/tar.go
@@ -109,6 +109,69 @@ func BasicInfoHeader(name string, size int64, fileInfo *winio.FileBasicInfo) *ta
 	return hdr
 }
 
+// SecurityDescriptorFromTarHeader reads the SDDL associated with the header of the current file
+// from the tar header and returns the security descriptor into a byte slice.
+func SecurityDescriptorFromTarHeader(hdr *tar.Header) ([]byte, error) {
+	// Maintaining old SDDL-based behavior for backward
+	// compatibility.  All new tar headers written by this library
+	// will have raw binary for the security descriptor.
+	var sd []byte
+	var err error
+	if sddl, ok := hdr.PAXRecords[hdrSecurityDescriptor]; ok {
+		sd, err = winio.SddlToSecurityDescriptor(sddl)
+		if err != nil {
+			return nil, err
+		}
+	}
+	if sdraw, ok := hdr.PAXRecords[hdrRawSecurityDescriptor]; ok {
+		sd, err = base64.StdEncoding.DecodeString(sdraw)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return sd, nil
+}
+
+// ExtendedAttributesFromTarHeader reads the EAs associated with the header of the
+// current file from the tar header and returns it as a byte slice.
+func ExtendedAttributesFromTarHeader(hdr *tar.Header) ([]byte, error) {
+	var eas []winio.ExtendedAttribute
+	var eadata []byte
+	var err error
+	for k, v := range hdr.PAXRecords {
+		if !strings.HasPrefix(k, hdrEaPrefix) {
+			continue
+		}
+		data, err := base64.StdEncoding.DecodeString(v)
+		if err != nil {
+			return nil, err
+		}
+		eas = append(eas, winio.ExtendedAttribute{
+			Name:  k[len(hdrEaPrefix):],
+			Value: data,
+		})
+	}
+	if len(eas) != 0 {
+		eadata, err = winio.EncodeExtendedAttributes(eas)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return eadata, nil
+}
+
+// EncodeReparsePointFromTarHeader reads the ReparsePoint structure from the tar header
+// and encodes it into a byte slice. The file for which this function is called must be a
+// symlink.
+func EncodeReparsePointFromTarHeader(hdr *tar.Header) []byte {
+	_, isMountPoint := hdr.PAXRecords[hdrMountPoint]
+	rp := winio.ReparsePoint{
+		Target:       filepath.FromSlash(hdr.Linkname),
+		IsMountPoint: isMountPoint,
+	}
+	return winio.EncodeReparsePoint(&rp)
+}
+
 // WriteTarFileFromBackupStream writes a file to a tar writer using data from a Win32 backup stream.
 //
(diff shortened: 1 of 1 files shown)


backuptar/tar.go
```diff
@@ -109,6 +109,72 @@ func BasicInfoHeader(name string, size int64, fileInfo *winio.FileBasicInfo) *ta
...
+			return nil, err
+		}
+	}
+	return eadata, nil
+}
+
+// EncodeReparsePointFromTarHeader checks if the given file is a symlink. If yes, then it
+// reads the ReparsePoint structure from the tar header and encodes it into a byte
+// array. If the file is not a symlink then returns an empty byte array.
+func EncodeReparsePointFromTarHeader(hdr *tar.Header) []byte {
```
dcantah, line 166: This only cares about (and handles) symlinks so the name is a bit awkward. Would `EncodeSymlinkFromTarHeader` be preferred?

backuptar/tar.go
```diff
@@ -392,13 +434,9 @@ func WriteBackupStreamFromTarFile(w io.Writer, t *tar.Reader, hdr *tar.Header) (
 			return nil, err
 		}
 	}
+
 	if hdr.Typeflag == tar.TypeSymlink {
```
dcantah, line 438: This check seems redundant now that `EncodeReparsePointFromTarHeader` also performs it (albeit in reverse)

backuptar/tar.go
```diff
@@ -109,6 +109,72 @@ func BasicInfoHeader(name string, size int64, fileInfo *winio.FileBasicInfo) *ta
 	return hdr
 }
 
+// EncodeSDDLFromTarHeader reads the SDDL associated with the header of the current file
+// from the tar header and encodes it into a byte array.
```
dcantah, line 113: nit: byte slice*

backuptar/tar.go
```diff
@@ -109,6 +109,72 @@ func BasicInfoHeader(name string, size int64, fileInfo *winio.FileBasicInfo) *ta
...
+		sd, err = base64.StdEncoding.DecodeString(sdraw)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return sd, nil
+}
+
+// EncodeExtendedAttributesFromTarHeader reads the EAs associated with the header of the current file
+// from the tar header and encodes it into a byte array.
```
dcantah, line 136: nit: byte slice*

backuptar/tar.go
```diff
@@ -109,6 +109,72 @@ func BasicInfoHeader(name string, size int64, fileInfo *winio.FileBasicInfo) *ta
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return eadata, nil
+}
+
+// EncodeReparsePointFromTarHeader checks if the given file is a symlink. If yes, then it
+// reads the ReparsePoint structure from the tar header and encodes it into a byte
+// array. If the file is not a symlink then returns an empty byte array.
```
dcantah, line 165: nit: byte slice*

ambarve: fixed.

backuptar/tar.go
ambarve: fixed.

backuptar/tar.go
ambarve: fixed.

backuptar/tar.go
ambarve: PTAL.

backuptar/tar.go
ambarve: I think it is not that this function _only_ handles symlinks. Rather for a file to be a valid reparse point its type should be symlink. So that type check for symlink is more of a verification.
Anyway, I have removed that check from this function since it is already done before calling this function. Let me know what you think about the new definition.

dcantah: My comment could've been worded better, but yes this makes it more apparent what it does and what's needed, thanks! Possibly add in what someone would need to check for on the header to verify they meet this requirement also. e.g. "...must be a symlink (`hdr.Typeflag == tar.TypeSymlink`)"

dcantah (approved):

backuptar/tar.go
```diff
@@ -109,6 +109,69 @@ func BasicInfoHeader(name string, size int64, fileInfo *winio.FileBasicInfo) *ta
 	return hdr
 }
 
+// EncodeSDDLFromTarHeader reads the SDDL associated with the header of the current file
+// from the tar header and encodes it into a byte slice.
+func EncodeSDDLFromTarHeader(hdr *tar.Header) ([]byte, error) {
```
kevpar, line 114: This is really returning a security descriptor, not SDDL. So maybe `SecurityDescriptorFromTarHeader` would be better.

kevpar:
Perhaps in the future we should have a separate package concerned with encoding Windows files in a tar, and `backuptar` should just be about interfacing between that and backup read/write.

backuptar/tar.go
```diff
@@ -109,6 +109,69 @@ func BasicInfoHeader(name string, size int64, fileInfo *winio.FileBasicInfo) *ta
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return sd, nil
+}
+
+// EncodeExtendedAttributesFromTarHeader reads the EAs associated with the header of the
+// current file from the tar header and encodes it into a byte slice.
+func EncodeExtendedAttributesFromTarHeader(hdr *tar.Header) ([]byte, error) {
```
kevpar, line 137: Prefixing these functions with `Encode` seems a bit odd to me, especially since we are really parsing or decoding data from the tar header. What do you think about just dropping `Encode` from the names (e.g. `ExtendedAttributesFromTarHeader`)?

kevpar:
Would like to see a commit title that is clearer on what this does. I would suggest something like `backuptar: Export tar header parsing functions`.

backuptar/tar.go
ambarve: I had the `Encode` prefix because we return the encoded attributes or security descriptors but I agree that the primary job of this function is not encoding things. I have dropped the prefix.

backuptar/tar.go
ambarve: Fixed.

ambarve:
@kevpar  I have updated the commit message. Can you please take another look?

kevpar (approved):
LGTM

PR #226, opened 2021-12-01
Fix 'OpenVirtualDiskParameters' BOOL fields

dcantah:
While [reworking the vhd package](https://github.com/microsoft/go-winio/commit/e9604c63ae982883093e4d3c33867ce24416af05#diff-e1200af476f67398dde7575bf9966bdccaf140ce444569ff78c485a43f48957a) I'd mistakingly replaced some of the fields
in the OpenVersion2 structure with the incorrect types. They're defined as
Windows BOOLS which is just a type alias for an int, and I'd put their type as
go bools.

As the type is already passed into some functions, to avoid a breaking change just
convert the incorrect type to a new structure with the correct definition internally.
Truthfully this turns out a bit better as supplying a bool makes more sense and is
more go friendly.

### DIFF
--- a/vhd/vhd.go
+++ b/vhd/vhd.go
@@ -1,3 +1,4 @@
+//go:build windows
 // +build windows
 
 package vhd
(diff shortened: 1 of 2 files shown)


vhd/vhd.go
```diff
@@ -1,3 +1,4 @@
+//go:build windows
 // +build windows
```
ambarve, line 2: Do we need both tags?

vhd/vhd.go
```diff
@@ -62,13 +63,27 @@ type OpenVirtualDiskParameters struct {
...
+}
+
 type AttachVersion2 struct {
 	RestrictedOffset uint64
 	RestrictedLength uint64
 }
 
 type AttachVirtualDiskParameters struct {
-	Version  uint32 // Must always be set to 2
+	Version  uint32
```
ambarve, line 86: Was this comment incorrect?

vhd/vhd.go
dcantah: gofmt will automatically apply the go 1.17 syntax of build tags to anything if you're on 1.17. Our go.mod here lists 1.12 as the earliest go version that you could build this project with, so we'd have to have both for the time being. However, what I've seen from perusing around yee olde github (https://cs.github.com/?q=%22%2F%2Fgo%3Abuild%22) is the overwhelming majority of folks even if they updated their go directive in their go.mod to be 1.17 is they've kept both syntaxes.

vhd/vhd.go
dcantah: Yea. From checking the OS, version 2 wasn't there on RS5 so it was a bit misleading.

ambarve (approved):
LGTM

vhd/vhd.go
```diff
@@ -62,13 +63,27 @@ type OpenVirtualDiskParameters struct {
 	Version2 OpenVersion2
 }
 
+// The OpenVersion2 structure was mistakingly changed to have the GetInfoOnly and ReadOnly fields be go bools instead
+// of int32's (they're defined as Windows BOOL's so int32 is the correct type). Fix this mistake here by making the right structure we need to pass
+// and not exporting it. We'll translate to the correct type as needed in any of the Open___ calls.
```
katiewasnothere, line 68: I'm fine with this PR and I think it does make more sense to use bools for higher-level calls, but I don't think we should refer to this as "fixing the mistake" in the comment. 

Instead we could just say 
```suggestion
// The higher level `OpenVersion2` struct uses bools to refer to `GetInfoOnly` and `ReadOnly` for ease of use. However,
// the internal windows structure uses `BOOLS` aka int32s for these types. `openVersion2` is used for translating 
// `OpenVersion2` fields to the correct windows internal field types on `Open____` calls. 
```

dcantah: I'd worded it like this as a lot of the calls expose a pretty faithful representation of all of the fields for the underlying windows API so this felt out of place with the rest. But we can mask it as us being go friendly 😆

dcantah: It works out well in the end

dcantah: Changed

dcantah:
@katiewasnothere PTAL

katiewasnothere (approved):
lgtm

PR #227, opened 2021-12-01
Replace github.com/pkg/errors with stdlib errors

bitgestalt:
Since Go version 1.13, the standard library has built-in support for
wrapping and unwrapping of error values.  This commit bumps the minimum
version required by the module from 1.12 to 1.13, replaces all calls to
errors.Wrap/f with fmt.Errorf and removes the now unneeded dependency on
github.com/pkg/errors.

Signed-off-by: Michael Hofmann <michael.hofmann@bitgestalt.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,9 +1,8 @@
 module github.com/Microsoft/go-winio
 
-go 1.12
+go 1.13
 
 require (
-	github.com/pkg/errors v0.9.1
 	github.com/sirupsen/logrus v1.7.0
 	golang.org/x/sys v0.0.0-20210124154548-22da62e12c0c
 )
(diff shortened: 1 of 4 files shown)


katiewasnothere:
@bitgestalt Was there a specific need for this? I think we prefer using the `errors` package here.

bitgestalt:
> @bitgestalt Was there a specific need for this? I think we prefer using the `errors` package here.

`errors` is a third-party package and, insofar as it is used by this module, its functionality is fully covered by the Go standard library. On top of that, `errors` is currently in maintenance mode and no longer actively under development and apparently also [unmaintained](https://github.com/pkg/errors/issues/245) . The one feature `errors.Wrap` has and `fmt.Errorf` has not - stack traces on errors - is only marginally useful since only a small subset of errors returned by functions in this module are created by `errors.Wrap`.

In short, unless you want to maintain strict compatibility with Go 1.12, I can see no good reason to keep depending on the external `errors` package.

katiewasnothere (approved):
Ah okay, seems like a bunch of projects are doing this as well. I'm fine with it then.

dcantah (approved):

PR #228, opened 2021-12-03
Add lookup of account names by SID

### LINKED ISSUE #202: Test case fails because of localized account names
[TestLookupValidSid](https://github.com/microsoft/go-winio/blob/5516f17a595844b6011597b41d9573aafabb6b6a/sd_test.go#L16) tries to lookup the SID for the group "Everyone". Since Windows uses localized names for built-in accounts, this test case will fail when executed under non-English versions of Windows.

bitgestalt:
This commit adds a new function `LookupNameBySid` that wraps
`LookupAccountSidW` and works as an inverse of the already existing
`LookupSidByName`. In addition to offering new functionality, this
fixes #202.

Signed-off-by: Michael Hofmann <michael.hofmann@bitgestalt.com>


sd.go
```diff
@@ -73,6 +78,38 @@ func LookupSidByName(name string) (sid string, err error) {
...
+		return "", &AccountLookupError{sid, cERROR_NONE_MAPPED}
+	}
+
+	sidBuffer, err := syscall.UTF16FromString(sid)
+	if err != nil {
+		return "", &AccountLookupError{sid, err}
+	}
+
+	var sidPtr *byte
+	convertStringSidToSid(&sidBuffer[0], &sidPtr)
```
dcantah, line 93: Is the pointer to the first elem needed? I'd think you could just pass sidBuffer directly. Also seems like this function can fail so can we add an (err error) to the //sys definition? https://github.com/golang/sys/blob/1c1b9b1eba6a51cf62024505e66017fcd6aa559f/windows/security_windows.go#L187

sd.go
```diff
@@ -73,6 +78,38 @@ func LookupSidByName(name string) (sid string, err error) {
...
+
+	nameBuffer := make([]uint16, nameSize)
+	refDomainBuffer := make([]uint16, refDomainSize)
+	err = lookupAccountSid(nil, sidPtr, &nameBuffer[0], &nameSize, &refDomainBuffer[0], &refDomainSize, &sidNameUse)
+	if err != nil {
+		return "", &AccountLookupError{sid, err}
+	}
+
+	name = syscall.UTF16ToString(nameBuffer)
+	return name, nil
```
dcantah, line 110:
```suggestion
	return windows.UTF16ToString(nameBuffer), nil
```

sd.go
```diff
@@ -73,6 +78,38 @@ func LookupSidByName(name string) (sid string, err error) {
 	return sid, nil
 }
 
+// LookupNameBySid looks up the name of an account by SID
+func LookupNameBySid(sid string) (name string, err error) {
+	if sid == "" {
+		return "", &AccountLookupError{sid, cERROR_NONE_MAPPED}
+	}
+
+	sidBuffer, err := syscall.UTF16FromString(sid)
```
dcantah, line 87: I'd prefer the x/sys/windows variants for any syscall usage in this function. I know we're not great about that in this project at the moment but we should switch over soon

dcantah:
Thanks for the change!

sd.go
bitgestalt: `sidBuffer` is a []uint16 and can't be passed to `convertStringSidToSid` directly. I have added an error check, though.

sd.go
bitgestalt: Fixed.

sd.go
bitgestalt: Fixed.

sd.go
dcantah: Ahh, I'm so used to the windows.UTF16PtrFromString variant I didn't even notice

bitgestalt: I had to double-check as well. All those pointer-slice conversions can get pretty confusing. :smiley:

dcantah: It'd be nice to swap to that function but not a blocker

dcantah (approved):
lgtm, small suggestion

sd.go
bitgestalt: OK, I've just done that.

dcantah:
@ambarve ptal

ambarve (approved):
One minor comment but otherwise LGTM!

PR #229, opened 2022-01-10
Add Getter Setters for Extended Attributes and Reparse points.

ambarve:
This PR adds the Go wrappers for getting / setting the extended file attributes and reparse points. This is required for preparing the container scratch directly from hcsshim instead of calling into existing OS APIs.


reparse.go
```diff
@@ -126,3 +138,83 @@ func EncodeReparsePoint(rp *ReparsePoint) []byte {
...
+	return buf.Bytes()
+}
+
+// Decode decodes the byte array (returned by GetReparsePoint) into
+// the current ReparseDataBuffer struct.
+func (r *ReparseDataBuffer) Decode(data []byte) error {
+	rdr := bytes.NewReader(data)
+	err := binary.Read(rdr, binary.LittleEndian, r)
+	if err != nil {
+		return errors.Wrap(err, "failed to decode reparse data buffer")
```
dcantah, line 175: There was actually a recent change that aimed at getting rid of the github.com/pkg/errors package in this repo 😆 https://github.com/microsoft/go-winio/pull/227/files. So best to swap to the stdlib form here and for the other wraps throughout

ea.go
```diff
@@ -13,6 +18,9 @@ type fileFullEaInformation struct {
 	ValueLength     uint16
 }
 
+//sys getFileEA(handle syscall.Handle, iosb *ioStatusBlock, buf *uint8, bufLen uint32, returnSingleEntry bool, eaList uintptr, eaListLen uint32, eaIndex *uint32, restartScan bool) (status ntstatus) = ntdll.NtQueryEaFile
+//sys setFileEA(handle syscall.Handle, iosb *ioStatusBlock, buf *uint8, bufLen uint32) (status ntstatus) = ntdll.NtSetEaFile
+
```
dcantah, line 23: For new bindings we should probably prefer `windows.Handle`. I see you have `GetFileEAHandle` defined to take in a `windows.Handle` also so it'll save the unnecessary cast when calling into these

ea.go
```diff
@@ -28,6 +36,69 @@ type ExtendedAttribute struct {
 	Flags uint8
 }
 
+func GetFileEA(path string) ([]ExtendedAttribute, error) {
```
dcantah, line 39: Comments on the exported methods for go doc

ea.go
```diff
@@ -28,6 +36,69 @@ type ExtendedAttribute struct {
...
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file: %s", err)
+	}
+	defer f.Close()
+
+	return GetFileEAHandle(windows.Handle(f.Fd()))
+}
+
+func GetFileEAHandle(handle windows.Handle) ([]ExtendedAttribute, error) {
```
dcantah, line 49: With the way the *Handle methods are named it seems like I'd expect to get back some handle to the EA's themselves which doesn't make sense (guess this is the way my brain works haha). Is exposing string and handle equivalents just to make this as flexible as possible?

ea.go
```diff
@@ -28,6 +36,69 @@ type ExtendedAttribute struct {
...
+	return SetFileEAHandle(fileHandle, attrs)
+}
+
+func SetFileEAHandle(fileHandle windows.Handle, attrs []ExtendedAttribute) error {
+	encodedEA, err := EncodeExtendedAttributes(attrs)
+	if err != nil {
+		return fmt.Errorf("failed to encoded extended attributes: %w", err)
+	}
+
+	// default buffer size to start with
```
dcantah, line 92: Was this meant to go somewhere else?

ea_test.go
```diff
@@ -89,3 +98,40 @@ func Test_SetFileEa(t *testing.T) {
 		t.Fatalf("NtSetEaFile failed with %08x", r)
 	}
 }
+
+func Test_SetGetFileEA(t *testing.T) {
+	tempDir := t.TempDir()
```
dcantah, line 103: I keep forgetting this exists now, awesome 😁

reparse_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// retrieve reparse buffer for the link
+	reparseBuffer, err := GetReparsePoint(linkfilePath)
+	if err != nil {
+		t.Fatalf("failed to get reparse point: %s", err)
+	}
+
+	rp, err := DecodeReparsePoint(reparseBuffer)
+	if err != nil {
+		t.Logf("failed to decode reparse point: %s", err)
```
helsaawy, line 35: Should this be `Fatalf` if the parsing failed?

helsaawy (approved):

dcantah:
What will this be used for? A PR description and commit descriptions will be helpful

ea.go
kevpar: Do we really need the shortcut methods that take a file path? It's probably not too onerous to expect callers to pass a handle.

dcantah: Right, I think dropping the Handle off the names and having these be the only methods makes sense to me at least. Not the only method we have in this project that just takes in a handle and has no shortcut string equivalent https://github.com/microsoft/go-winio/blob/master/file.go#L113

reparse.go
```diff
@@ -126,3 +138,83 @@ func EncodeReparsePoint(rp *ReparsePoint) []byte {
...
+	return nil
+}
+
+// GetReparsePoint returns a byte array that represents a ReparseDataBuffer if
+// the file at `path` is a valid reparse point.
+func GetReparsePoint(path string) ([]byte, error) {
+	utf16DestPath := windows.StringToUTF16(path)
+	// We need to open the file with windows specific permissions so just using
+	// os.Open won't work.
+	fileHandle, err := windows.CreateFile(&utf16DestPath[0], (windows.GENERIC_READ | windows.GENERIC_WRITE), 0, nil, windows.OPEN_EXISTING, (windows.FILE_ATTRIBUTE_NORMAL | windows.FILE_FLAG_OPEN_REPARSE_POINT), 0)
```
msscotb, line 186: Probably don't need GENERIC_WRITE

reparse.go
```diff
@@ -126,3 +138,83 @@ func EncodeReparsePoint(rp *ReparsePoint) []byte {
...
+	return outBuf[:outBufLen], nil
+}
+
+// SetReparsePoint sets a reparse point at `path` with `data`. `data` should be
+// an encoded ReparseDataBuffer struct.
+func SetReparsePoint(path string, data []byte) error {
+	utf16DestPath := windows.StringToUTF16(path)
+	// We need to open the file with windows specific permissions so just using
+	// os.Open won't work.
+	fileHandle, err := windows.CreateFile(&utf16DestPath[0], (windows.GENERIC_READ | windows.GENERIC_WRITE), 0, nil, windows.CREATE_NEW, windows.FILE_ATTRIBUTE_NORMAL, 0)
```
msscotb, line 208: Probably don't need GENERIC_READ, maybe FILE_READ_ATTRIBUTES if that.

ea.go
ambarve: I added the method which takes the file path mainly for convenience and also so that we can encapsulate correct file opening permissions (for example, you need to open the file with specific access right when gettting/setting EA, or when gettting/setting reparse points). However, there are times when you already have a handle open for a file and you just want to pass in that (probably because you opened it in non sharing mode) that's why I added the other method. I agree that the handle methods can be named better but I think keeping both can be helpful. What do you think?

ea.go
ambarve: A leftover comment from refactoring.

reparse.go
ambarve: Fixed.

ea.go
```diff
@@ -28,6 +36,73 @@ type ExtendedAttribute struct {
 	Flags uint8
 }
 
+// GetFileEA retrieves the extended attributes for the file at `path`
+func GetFileEA(path string) ([]ExtendedAttribute, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file: %s", err)
```
dcantah, line 43: Prefer to wrap the error, %w

ea.go
```diff
@@ -28,6 +36,73 @@ type ExtendedAttribute struct {
...
+
+// SetFileEA sets the extended attributes for file at `path`.
+func SetFileEA(path string, attrs []ExtendedAttribute) error {
+	utf16Path := windows.StringToUTF16Ptr(path)
+
+	// We need to open the file with windows specific permissions so just using os.Open won't work.
+	fileAccessRightWriteEA := 0x10
+	fileHandle, err := windows.CreateFile(utf16Path, uint32(fileAccessRightWriteEA), 0, nil, windows.OPEN_EXISTING, windows.FILE_ATTRIBUTE_NORMAL, 0)
+	if err != nil {
+		return fmt.Errorf("open file failed with: %s", err)
```
dcantah, line 83: Wrap error

reparse.go
```diff
@@ -126,3 +137,83 @@ func EncodeReparsePoint(rp *ReparsePoint) []byte {
...
+	return nil
+}
+
+// GetReparsePoint returns a byte array that represents a ReparseDataBuffer if
+// the file at `path` is a valid reparse point.
+func GetReparsePoint(path string) ([]byte, error) {
+	utf16DestPath := windows.StringToUTF16(path)
+	// We need to open the file with windows specific permissions so just using
+	// os.Open won't work.
+	fileHandle, err := windows.CreateFile(&utf16DestPath[0], windows.GENERIC_READ, 0, nil, windows.OPEN_EXISTING, (windows.FILE_ATTRIBUTE_NORMAL | windows.FILE_FLAG_OPEN_REPARSE_POINT), 0)
```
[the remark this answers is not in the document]
ambarve: Good catch. Fixed and added a test.

ea.go
ambarve: fixed.

ea.go
ambarve: @kevpar / @dcantah  any thoughts on this?

dcantah: What would you propose for the new names for the handle variants? I'm still leaning on just having the handle functions only. If the operations require specific permissions when opening the file, that can be called out in the comment for the function ("file handle must have X | Y | Z permissions for this function" or similar).

ambarve: I was thinking something like `GetEAforHandle` and `GetFileEA`. I am okay with having handle functions only but then everytime you call them you will have to get a handle to that file by doing `windows.CreateFile` calls with specific permissions. Anyway, I will remove the file functions for now and keep only the handle functions. We can add file functions later if required.

ambarve: I have updated the getter setter functions for EA to only take the handle now. For reparse points I think it makes more sense to directly take filepath as the input instead of taking the handle so I have kept those functions as it is. For the get security descriptor function there is not much value in adding that function if the caller also has to open the handle themselves, so I have kept that function too. Let me know what you guys think.

ea.go
```diff
@@ -28,6 +35,49 @@ type ExtendedAttribute struct {
...
+	var iosb ioStatusBlock
+	// keep increasing the buffer size until it is large enough
+	for {
+		status := getFileEA(handle, &iosb, &buf[0], uint32(bufLen), false, 0, 0, nil, true)
+		if status < 0 {
+			// convert ntstatus code to windows error
+			werr := rtlNtStatusToDosError(status)
+			if werr == syscall.ERROR_INSUFFICIENT_BUFFER || werr == syscall.ERROR_MORE_DATA {
+				bufLen *= 2
+				buf = make([]byte, bufLen*2)
```
msscotb, line 53: Intended to quadruple buffer?

ea.go
msscotb: If our use case for these is to open a handle, and then query SDs, EAs and reparse points it make sense to just have the NtAPI style that take handles. Otherwise there is a separate createfile on each function call.

ea.go
```diff
@@ -28,6 +35,49 @@ type ExtendedAttribute struct {
...
+			} else {
```
msscotb, line 54: Doesn't the output of the syscall give the exact size required for the buffer?

ambarve: This particular syscall doesn't seem to return the size of the required buffer. (https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-zwqueryeafile)

ea.go
ambarve: Fixed.

ea.go
ambarve: I think for reparse points its best to have the functions that take file paths but for EA we mostly deal with the handles so I have kept handles there. For SDs there is no particular preference but the wrapper from `golang.org/x/sys/windows` already  provides a function which takes in the handle so I kept the file path version here.

msscotb: Can you explain this comment a bit more? Why is it best to take paths for reparsepoint functions?

ea.go
msscotb: iosb.Information should contain the size of the required buffer

ea.go
ambarve: I only said that because in order to get handle to a reparse point you need to open it with the `FLAG_OPEN_REPARSE_POINT` flag so its better to do that in the call that gets or sets the reparse points than expecting the caller to do it.

ea.go
```diff
@@ -28,6 +35,49 @@ type ExtendedAttribute struct {
...
+	bufLen := 1024
+	buf := make([]byte, bufLen)
+	var iosb ioStatusBlock
+	// keep increasing the buffer size until it is large enough
+	for {
+		status := getFileEA(handle, &iosb, &buf[0], uint32(bufLen), false, 0, 0, nil, true)
+		if status < 0 {
+			// convert ntstatus code to windows error
+			werr := rtlNtStatusToDosError(status)
+			if werr == syscall.ERROR_INSUFFICIENT_BUFFER || werr == syscall.ERROR_MORE_DATA {
```
dcantah, line 51: nit but x/sys/windows has these defined as well so would be preferred

ea.go
```diff
@@ -28,6 +35,49 @@ type ExtendedAttribute struct {
...
+// GetFileEA retrieves the extended attributes for the file represented by `handle`. The
+// `handle` must have been opened with file access flag FILE_READ_EA (0x8).
+func GetFileEA(handle windows.Handle) ([]ExtendedAttribute, error) {
+	// default buffer size to start with
+	bufLen := 1024
+	buf := make([]byte, bufLen)
+	var iosb ioStatusBlock
+	// keep increasing the buffer size until it is large enough
+	for {
+		status := getFileEA(handle, &iosb, &buf[0], uint32(bufLen), false, 0, 0, nil, true)
```
dcantah, line 47: The ntstatus type (that's returned from getFileEA) has an Err() method that does the status and converting to dos error inside https://github.com/microsoft/go-winio/blob/87532d1cfecaa30ec76bc0b281875615ffb20a8f/pipe.go#L61

ea.go
```diff
@@ -28,6 +35,49 @@ type ExtendedAttribute struct {
...
+func SetFileEA(handle windows.Handle, attrs []ExtendedAttribute) error {
+	encodedEA, err := EncodeExtendedAttributes(attrs)
+	if err != nil {
+		return fmt.Errorf("failed to encoded extended attributes: %w", err)
+	}
+
+	var iosb ioStatusBlock
+
+	status := setFileEA(handle, &iosb, &encodedEA[0], uint32(len(encodedEA)))
+	if status < 0 {
```
dcantah, line 75: Same ntstatus Err method comment as above. You could do 
```go
return setFileEA(handle, &iosb, &encodedEA[0], uint32(len(encodedEA))).Err()
```

sd_test.go
```diff
@@ -26,3 +28,23 @@ func TestLookupEmptyNameFails(t *testing.T) {
 		t.Fatalf("expected AccountLookupError with ERROR_NONE_MAPPED, got %s", err)
 	}
 }
+
+func TestGetFileSDDL(t *testing.T) {
+	win32 := "C:\\Windows\\System32"
+	kern := "C:\\Windows\\System32\\kernel32.dll"
```
dcantah, line 34: Can use raw string literals to avoid the escaping 
```go
win32 := `C:\windows\system32`
```

fileinfo.go
```diff
@@ -36,6 +36,16 @@ func SetFileBasicInfo(f *os.File, bi *FileBasicInfo) error {
 	return nil
 }
 
+// SetFileBasicInfoForHandle sets times and attributes for a file represented by
+// the given handle
+func SetFileBasicInfoForHandle(f windows.Handle, bi *FileBasicInfo) error {
+	if err := windows.SetFileInformationByHandle(f, windows.FileBasicInfo, (*byte)(unsafe.Pointer(bi)), uint32(unsafe.Sizeof(*bi))); err != nil {
+		return &os.PathError{Op: "SetFileInformationByHandle", Err: err}
+	}
+	runtime.KeepAlive(f)
```
dcantah, line 45: Trying to wrap my head around if we actually need this for a uintptr

dcantah:
I almost forgot we have a Linux CI here now haha, but it's not happy with this PR apparently.

ea.go
msscotb: Ah right the Information field is only populated on success with the count of bytes read but not on failure.

ea.go
ambarve: Fixed

ea.go
ambarve: @msscotb , @kevpar , @dcantah I have updated the PR, with the latest set of changes getters/setters for EAs take a file handle, getter for SDDL takes a path and getter for reparse takes a path but setter for reparse takes a file handle. The reason for this is that a reparse point could be created on a normal file or a directory so I have left the job of opening the handle to it on the caller. Let me know if this looks good.

ambarve:
> I almost forgot we have a Linux CI here now haha, but it's not happy with this PR apparently.

Why do we need Linux CI for this repo?

dcantah:
> > I almost forgot we have a Linux CI here now haha, but it's not happy with this PR apparently.
> 
> Why do we need Linux CI for this repo?

There's some code that can run on linux, so probably for sanity https://github.com/microsoft/go-winio/pull/241

reparse_test.go
```diff
@@ -0,0 +1,113 @@
+package winio
```
helsaawy, line 1: Can you add a `//go:build windows` constraint to this file and see if that resolves the Linux CI problem?

sd.go
```diff
@@ -97,3 +100,16 @@ func SecurityDescriptorToSddl(sd []byte) (string, error) {
 	defer localFree(uintptr(unsafe.Pointer(sddl)))
 	return syscall.UTF16ToString((*[0xffff]uint16)(unsafe.Pointer(sddl))[:]), nil
 }
+
+func GetFileSecurityDescriptor(path string) (*windows.SECURITY_DESCRIPTOR, error) {
+	utf16Path := windows.StringToUTF16(path)
```
dcantah, line 105: Prefer `windows.UTF16PtrFromString`

reparse.go
```diff
@@ -129,3 +140,69 @@ func EncodeReparsePoint(rp *ReparsePoint) []byte {
...
+	if err != nil {
+		fmt.Errorf("failed to decode reparse data buffer: %w", err)
+	}
+	return nil
+}
+
+// GetReparsePoint returns a byte array that represents a ReparseDataBuffer if
+// the file at `path` is a valid reparse point.
+func GetReparsePoint(path string) ([]byte, error) {
+	utf16DestPath := windows.StringToUTF16(path)
```
dcantah, line 185: Prefer `windows.UTF16PtrFromString`

reparse.go
```diff
@@ -129,3 +140,69 @@ func EncodeReparsePoint(rp *ReparsePoint) []byte {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get reparse point for file %s: %w", path, err)
+	}
+
+	return outBuf[:outBufLen], nil
+}
+
+// SetReparsePoint sets a reparse point with `data` on the file/directory represented by
+// `handle` . `data` should be an encoded ReparseDataBuffer struct.
+func SetReparsePoint(handle windows.Handle, data []byte) error {
```
dcantah, line 206: Would it make sense to make these "friendly" APIs and just take in ReparseDataBuffers directly and we encode inside and if folks want to call encode/decode and pass around byte slices themselves we leave those methods exported?

dcantah:
@ambarve We were looking to get out a new winio release soon for Hamza's hvsocket changes/fixes, lemme know if you think we'd be able to get this in

reparse.go
ambarve: IIRC the problem here is that we currently have two structs representing the reparse data buffer in this code. One is the `ReparsePoint` struct and other is the newly added `ReparseDataBuffer` struct. The `ReparsePoint` struct only supports symlinks and mount points while `ReparseDataBuffer` supports all kinds of reparse points. We have to keep the `ReparsePoint` struct for backwards compatibility. We want to use the `SetReparsePoint` call for a buffer that could be an encoded buffer of `ReparseDataBuffer` or `ReparsePoint`, hence the []byte array. Let me know what you think?

reparse.go
ambarve: I have changed this to `UTF16FromString`. I don't think we need to use `UTF16PtrFromString` but let me know if you think otherwise.

fileinfo.go
```diff
@@ -37,6 +37,15 @@ func SetFileBasicInfo(f *os.File, bi *FileBasicInfo) error {
 	return nil
 }
 
+// SetFileBasicInfoForHandle sets times and attributes for a file represented by
+// the given handle
+func SetFileBasicInfoForHandle(f windows.Handle, bi *FileBasicInfo) error {
+	if err := windows.SetFileInformationByHandle(f, windows.FileBasicInfo, (*byte)(unsafe.Pointer(bi)), uint32(unsafe.Sizeof(*bi))); err != nil {
+		return &os.PathError{Op: "SetFileInformationByHandle", Err: err}
```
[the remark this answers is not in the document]
ambarve: Changed to `SyscallError`.

ea.go
```diff
@@ -28,6 +37,44 @@ type ExtendedAttribute struct {
...
+		status := getFileEA(handle, &iosb, &buf[0], uint32(bufLen), false, 0, 0, nil, true)
+		if status.Err() != nil {
+			// convert ntstatus code to windows error
+			if status.Err() == windows.ERROR_INSUFFICIENT_BUFFER || status.Err() == windows.ERROR_MORE_DATA {
+				bufLen *= 2
+				buf = make([]byte, bufLen)
+			} else {
+				return nil, fmt.Errorf("get file EA failed with: %w", status.Err())
+			}
+		} else {
```
dcantah, line 58: Don't need the else

reparse.go
dcantah: Well the bit that confused me is the comment, it states that it should be an encoded ReparseDataBuffer explicitly, so I thought this didn't care about the old definition

reparse.go
```diff
@@ -129,3 +140,72 @@ func EncodeReparsePoint(rp *ReparsePoint) []byte {
...
```
dcantah, line 209: I still think the split here in API between Get taking in a path and Set using a handle is strange. Maybe we expose a method to call CreateFile with the right permissions you need to get/set and have these all take in handles. Or document what flags you need for CreateFile, but it seems like quite a few

ea.go
aneesh-n: We need the else with break otherwise the for loop doesn't end and the code gets stuck in a loop. 
Existing code is correct.

dcantah: I should've been more clear. Keep the break just don't need the else. But we'd need something to retry the loop so we'd need a continue after `buf = make([]byte, bufLen)`, e.g. This seems mostly preference
```go
for {
    status := getFileEA(handle, &iosb, &buf[0], uint32(bufLen), false, 0, 0, nil, true)
    if status.Err() != nil {
	// convert ntstatus code to windows error
	if status.Err() == windows.ERROR_INSUFFICIENT_BUFFER || status.Err() == windows.ERROR_MORE_DATA {
		bufLen *= 2
		buf = make([]byte, bufLen)
                continue
	} else {
                return nil, fmt.Errorf("get file EA failed with: %w", status.Err())
        }
   }
   break
}



```

aneesh-n: I see what you mean, that does seem a bit more concise.

ea.go
```diff
@@ -28,6 +37,44 @@ type ExtendedAttribute struct {
...
+// `handle` must have been opened with file access flag FILE_READ_EA (0x8).
+func GetFileEA(handle windows.Handle) ([]ExtendedAttribute, error) {
+	// default buffer size to start with
+	bufLen := 1024
+	buf := make([]byte, bufLen)
+	var iosb ioStatusBlock
+	// keep increasing the buffer size until it is large enough
+	for {
+		status := getFileEA(handle, &iosb, &buf[0], uint32(bufLen), false, 0, 0, nil, true)
+		if status.Err() != nil {
```
aneesh-n, line 50: Thank you for this contribution @ambarve and I'm thankful to all the contributors of this project, it's been very helpful.

I think one aspect that could be improved in this function is that if you pass a handle to a file which does not have any extended attributes, then it returns an error. However, this should ideally be a normal situation that should be handled gracefully.
I noticed that the system call returns -1073741742 status code for both files and directories in this case.

Suggested change -
for { status := getFileEA(handle, &iosb, &buf[0], uint32(bufLen), false, 0, 0, nil, true) if status == -1073741742 { //If status is -1073741742, no extended attributes were found return nil, nil } else { err := status.Err() if err != nil { // convert ntstatus code to windows error if err == windows.ERROR_INSUFFICIENT_BUFFER || err == windows.ERROR_MORE_DATA { bufLen *= 2 buf = make([]byte, bufLen) continue } return nil, fmt.Errorf("get file EA failed with: %w", err) } } break }

Said that, I do not know if this is the best way to handle this scenario or if that status code will always be consistent.
Anyone with a more comprehensive understanding of the system internals have any better suggestions?

ea_test.go
```diff
@@ -90,3 +101,48 @@ func Test_SetFileEa(t *testing.T) {
...
+	var readEAs []ExtendedAttribute
+	if readEAs, err = GetFileEA(fileHandle); err != nil {
+		t.Fatalf("get EA for file failed: %s", err)
+	}
+
+	if !reflect.DeepEqual(readEAs, testEAs) {
+		t.Logf("expected: %+v, found: %+v\n", testEAs, readEAs)
+		t.Fatalf("EAs read from testfile don't match")
+	}
+}
```
aneesh-n, line 148: Another useful test would be for setting and getting extended attributes for folders.
The main change required is the passing of an extra attribute - windows.FILE_FLAG_BACKUP_SEMANTICS

`

func Test_SetGetFolderEA(t *testing.T) {
	tempDir := t.TempDir()
	testfolderPath := filepath.Join(tempDir, "testfolder")
	// create temp folder
	err := os.Mkdir(testfolderPath, os.ModeDir)
	if err != nil {
		t.Fatalf("failed to create temporary file: %s", err)
	}

	nAttrs := 3
	testEAs := make([]ExtendedAttribute, 3)
	// generate random extended attributes for test
	for i := 0; i < nAttrs; i++ {
		// EA name is automatically converted to upper case before storing, so
		// when reading it back it returns the upper case name. To avoid test
		// failures because of that keep the name upper cased.
		testEAs[i].Name = fmt.Sprintf("TESTEA%d", i+1)
		testEAs[i].Value = make([]byte, rand.Int31n(math.MaxUint8))
		rand.Read(testEAs[i].Value)
	}

	utf16Path := windows.StringToUTF16Ptr(testfolderPath)
	fileAccessRightReadWriteEA := (0x8 | 0x10)
	fileHandle, err := windows.CreateFile(utf16Path, uint32(fileAccessRightReadWriteEA), 0, nil, windows.OPEN_EXISTING, windows.FILE_ATTRIBUTE_NORMAL|windows.FILE_FLAG_BACKUP_SEMANTICS, 0)

	if err != nil {
		t.Fatalf("open folder failed with: %s", err)
	}
	defer windows.Close(fileHandle)

	if err := SetFileEA(fileHandle, testEAs); err != nil {
		t.Fatalf("set EA for folder failed: %s", err)
	}

	var readEAs []ExtendedAttribute
	if readEAs, err = GetFileEA(fileHandle); err != nil {
		t.Fatalf("get EA for folder failed: %s", err)
	}

	if !reflect.DeepEqual(readEAs, testEAs) {
		t.Logf("expected: %+v, found: %+v\n", testEAs, readEAs)
		t.Fatalf("EAs read from test folder don't match")
	}
}

`

ambarve:
These changes are not required any more so closing this out. Will be reopened later if required.

PR #231, opened 2022-02-07
HvsockConn shutdown bugfix, added .IsClosed() functions

helsaawy:
Corrected `HvsockConn` `CloseRead`/`Write` to check if the socket
has been closed before attempting to shutdown reading or writing.
Additionally, `shutdown` was not respecting whether to shutdown reading
or writing, and only shutting down reading.

Added a `CloseReadWrite` function to `HvsockConn` to expose shutting
down both ends of the connection.

Added `IsClosed` function to `win32File` (and therefore `win32Pipe` and
`win32MessageBytePipe`) and `HvsockConn` to check if the socket/pipe
has been closed already.


ambarve:
Is this change fixing some bug? Can you also add some more context on that please?

helsaawy:
@ambarve I added more context, ptal

dcantah (approved):

hvsock.go
```diff
@@ -279,6 +288,15 @@ func (conn *HvsockConn) CloseWrite() error {
 	return nil
 }
 
+// CloseReadWrite shuts down both the read and write end of the socket, preventing future operations.
+func (conn *HvsockConn) CloseReadWrite() error {
```
ambarve, line 292: Is this added because it is being used somewhere in hcsshim? I am mainly asking because the `IsClosed` call will work correctly only when the `HvsockConn` is closed using the `Close` method. But if `CloseReadWrite` is called then `IsClosed` will keep returning false. Unless we have a specific reason to call `CloseReadWrite` instead of `Close` on a socket, let's avoid that.

helsaawy: I added it since the MS docs mention it as one of the options when calling shutdown, basically a combo `CloseRead` and `CloseWrite`. But I agree, it is confusing; Ill remove it.

ambarve (approved):
LGTM.

PR #233, opened 2022-02-18
backuptar: SecurityDescriptorFromTarHeader() don't decode twice

thaJeztah:
I noticed this when looking at #220

While writing, I wondered why we didn't just return the result of `base64.StdEncoding.DecodeString()` directly, but it looks like that may return "garbage" output on failure; https://github.com/golang/go/blob/go1.17.7/src/encoding/base64/base64.go#L382-L387

Here's a short example showing that; https://go.dev/play/p/6kCw85ofiTx

```go
package main

import (
	"encoding/base64"
	"fmt"
)

func main() {
	sd1, err := base64.StdEncoding.DecodeString("Hello, 世界")
	if err != nil {
		fmt.Println(err)
	}
	fmt.Printf("%+v: %s\n", sd1, string(sd1))
}
```

The above shows;

```
illegal base64 data at input byte 5
[29 233 101]: �e
```

While callers should never trust output if an error occurs, it's probably better to explicitly return `nil` just to be sure no garbage is used.

----

This may be a theoretical situation, but the tar headers may contain
both the old and new headers, in which case we would be decoding
both.

This patch rewrites the function to try the new headers first, and
return early if found, then fall back to trying the old headers.

### DIFF
--- a/backuptar/tar.go
+++ b/backuptar/tar.go
@@ -116,24 +116,23 @@ func BasicInfoHeader(name string, size int64, fileInfo *winio.FileBasicInfo) *ta
 // SecurityDescriptorFromTarHeader reads the SDDL associated with the header of the current file
 // from the tar header and returns the security descriptor into a byte slice.
 func SecurityDescriptorFromTarHeader(hdr *tar.Header) ([]byte, error) {
-	// Maintaining old SDDL-based behavior for backward
-	// compatibility.  All new tar headers written by this library
-	// will have raw binary for the security descriptor.
-	var sd []byte
-	var err error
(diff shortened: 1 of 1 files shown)


thaJeztah:
@ambarve @kevpar ptal

anmaxvl (approved):

dcantah (approved):
LGTM. @ambarve Still like you to take a quick peek as you're probably most familiar here

ambarve (approved):
LGTM

thaJeztah:
Thanks for reviewing, everyone!

PR #230, opened 2022-02-03
Bump go version to 1.17 in go.mod/CI

dcantah:
1.13 is sufficiently old at this point and net.ErrClosed from 1.16 will be nice
to use.

This change additionally runs go fmt to bring in the new 1.17 build tag syntax
and builds the binaries in our CI with 1.17+.

PR #235, opened 2022-02-23
Added HV Socket tests

helsaawy:
Added tests for Hyper-V sockets, including `CloseRead` and `CloseWrite`, and that addresses are set appropriately.

Bug fixes:
* Listen (and now Dial) now properly initialize their sockets correctly after `AcceptEx` (and `ConnectEx`), so that `GetSockName`, `GetPeerName`, and `Shutdown` (for `CloseRead` and `CloseWrite`) now function properly.
* The `socketError` used to check `bind`'s return value was incorrect, it should be `uintptr(^uint32(0))`,  not `^uintptr(0)`

Added resources to support testing:
* Well-know Hyper-V VMIDs for parents, children, and loopback.
* `Dial()` and `DialContext()` (along with a corresponding `HvsockDialer` struct) to dial a specific Hyper-V socket at an address.

Additionally, added functionality to [register an application](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#register-a-new-application) (service GUID and name) in the registry to enable Hyper-V sockets using that service GUID

Created a `sockets` package, currently only with syscalls to `Bind`, `ConnectEx` and `GetSockName`, bypassing `syscall/windows` restrictions that they only be used on socket address types defined by the package.

Upgraded to go v1.17 to use `unsafe.Slice()`.


hvsock.go
```diff
@@ -160,36 +271,146 @@ func (l *HvsockListener) Close() error {
...
+
+	// preemptive timeout/cancellation check
+	if err = ctx.Err(); err != nil {
+		return nil, conn.opErr(op, err)
+	}
+
 	sock, err := newHvSocket()
 	if err != nil {
-		return nil, err
+		return nil, conn.opErr("dial", err)
```
dcantah, line 312: Should use the `op` var we made on the first line

hvsock.go
```diff
@@ -160,36 +271,146 @@ func (l *HvsockListener) Close() error {
...
 	defer sock.wg.Done()
 	var bytes uint32
-	err = windows.ConnectEx(windows.Handle(sock.handle), sa, nil, 0, &bytes, &c.o)
-	_, err = sock.asyncIo(ctx, c, nil, bytes, err)
+	n := 1 + int(d.Retries)
+	for i := 1; i <= n; i++ {
+		err = sockets.ConnectEx(
+			windows.Handle(sock.handle), &sa,
+			nil /*sendBuf*/, 0, /*sendDataLen*/
+			&bytes, (*windows.Overlapped)(unsafe.Pointer(&c.o)))
```
dcantah, line 337: I'm not sure what the idiomatic answer is here, but I generally feel like every argument on a new line is easier to read. Also I'd prefer // comments to the c block comment style

pkg/sockets/sockets.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/Microsoft/go-winio/pkg/guid"
+	"golang.org/x/sys/windows"
+)
+
+//go:generate go run golang.org/x/sys/windows/mkwinsyscall -output zsyscall_windows.go sockets.go
+
+//sys getsockname(s windows.Handle, name unsafe.Pointer, namelen *int32) (err error) [failretval==socketError] = ws2_32.getsockname
+//sys getpeername(s windows.Handle, name unsafe.Pointer, namelen *int32) (err error) [failretval==socketError] = ws2_32.getpeername
+//sys bind(s windows.Handle, name unsafe.Pointer, namelen int32) (err error) [failretval==socketError] = ws2_32.bind
```
dcantah, line 18: All of these look to be in golang.org/x/sys/windows

pkg/sockets/sockets.go
```diff
@@ -0,0 +1,170 @@
...
+		return err
+	}
+
+	return bind(s, ptr, l)
+}
+
+// "golang.org/x/sys/windows".ConnectEx and .Bind only accept internal implementations of the
+// their sockaddr interface, so they cannot be used with HvsockAddr
+// Replicate functionality here from
+// https://cs.opensource.google/go/x/sys/+/master:windows/syscall_windows.go
```
dcantah, line 92: Ahh, this may be why you defined bind above.

dcantah:
This does a heck of a lot more than the title leads on 😆. Preferably this should be split across multiple commits (possibly even multiple PRs). I don't think we should bump to 1.17 in this PR at the very least, we can shoot to get this in https://github.com/microsoft/go-winio/pull/230, but I think the more important fact is the dial and registry work here

helsaawy:
>  I don't think we should bump to 1.17 in this PR at the very least, we can shoot to get this in #230, but I think the more important fact is the dial and registry work here

I think my implementation of dial needs 1.17 for unsafe.Slice, so I'll wait for #230 to be merged

helsaawy:
Ill split this into a functionality PR and a tests PR

PR #238, opened 2022-03-23
HVsock Registry Application Registration

helsaawy:
Using Hyper-V sockets requires [registering the application](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#register-a-new-application) in the Windows Registry.

This PR adds (de)registration functionality for an application name/GUID.

This PR is split from #235 

### DIFF
--- a/hvsock.go
+++ b/hvsock.go
@@ -12,6 +12,8 @@ import (
 	"time"
 	"unsafe"
 
+	"golang.org/x/sys/windows/registry"
+
 	"github.com/Microsoft/go-winio/pkg/guid"
 )
 
@@ -313,4 +315,32 @@ func (conn *HvsockConn) SetReadDeadline(t time.Time) error {
 // SetWriteDeadline implements the net.Conn SetWriteDeadline method.
 func (conn *HvsockConn) SetWriteDeadline(t time.Time) error {
 	return conn.sock.SetWriteDeadline(t)
+}
+
+// HvsockRegisterService registers the application defined by the guid and name with the
+// Hyper-V Host's registry. If the GUID is already registered, then the name is overwritten with
+// the new name.
+//
+// See: https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#register-a-new-application
+func HvsockRegisterService(id guid.GUID, name string) error {
+	k, _, err := registry.CreateKey(registry.LOCAL_MACHINE, hvSocketRegKey(id), registry.WRITE)
+	if err != nil {
+		return fmt.Errorf("could not create registry for %s: %w", id.String(), err)
+	}
+	defer k.Close()
+
+	if err = k.SetStringValue("ElementName", name); err != nil {
+		return fmt.Errorf("could not set service name to %s: %w", name, err)
+	}
+
+	return nil
+}
+
+// HvSockUnregisterService deleted the registration defined by the guid from the Hyper-V Host's registry.
+func HvsockUnregisterService(id guid.GUID) error {
+	return registry.DeleteKey(registry.LOCAL_MACHINE, hvSocketRegKey(id))
+}
+
+func hvSocketRegKey(id guid.GUID) string {
+	return `SOFTWARE\Microsoft\Windows NT\CurrentVersion\Virtualization\GuestCommunicationServices\` + id.String()
 }
--- a/hvsock.go
+++ b/hvsock.go
@@ -336,7 +336,7 @@ func HvsockRegisterService(id guid.GUID, name string) error {
 	return nil
 }
 
-// HvSockUnregisterService deleted the registration defined by the guid from the Hyper-V Host's registry.
+// HvsockUnregisterService deletes the registration defined by the guid from the Hyper-V Host's registry.
 func HvsockUnregisterService(id guid.GUID) error {
 	return registry.DeleteKey(registry.LOCAL_MACHINE, hvSocketRegKey(id))
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
+.vscode/
+
 *.exe


hvsock.go
```diff
@@ -314,3 +316,33 @@ func (conn *HvsockConn) SetReadDeadline(t time.Time) error {
 func (conn *HvsockConn) SetWriteDeadline(t time.Time) error {
 	return conn.sock.SetWriteDeadline(t)
 }
+
+// HvSockRegisterService registers the application defined by the guid and name with the
+// Hyper-V Host's registry.
+//
+// See: https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#register-a-new-application
+func HvSockRegisterService(id guid.GUID, name string) error {
```
dcantah, line 324: Can call out if the key already exists this is just a no-op

kevpar, line 324: Existing naming scheme seems to be `HvsockFoo` rather than `HvSockFoo`

hvsock.go
```diff
@@ -314,3 +316,33 @@ func (conn *HvsockConn) SetReadDeadline(t time.Time) error {
...
+	k, exists, err := registry.CreateKey(registry.LOCAL_MACHINE, hvSocketRegKey(id), registry.WRITE)
+	if err != nil {
+		return fmt.Errorf("could not create registry for %s: %w", id.String(), err)
+	}
+	defer k.Close()
+
+	if exists {
+		return nil
```
kevpar, line 332: If it already exists should we change the name (or check the name and error if different) instead of returning?

hvsock.go
helsaawy: Added that it overwrites the name if the id is registered

hvsock.go
```diff
@@ -314,3 +316,31 @@ func (conn *HvsockConn) SetReadDeadline(t time.Time) error {
...
+	defer k.Close()
+
+	if err = k.SetStringValue("ElementName", name); err != nil {
+		return fmt.Errorf("could not set service name to %s: %w", name, err)
+	}
+
+	return nil
+}
+
+// HvSockUnregisterService deleted the registration defined by the guid from the Hyper-V Host's registry.
```
dcantah, line 339:
```suggestion
// HvsockUnregisterService deletes the registration defined by the guid from the Hyper-V Host's registry.
```

dcantah (approved):

kevpar:
Do we know what precisely this registration affects or is needed for? I'm pretty sure we don't do this for any of the hvsocket stuff we use in the GCS today.

helsaawy:
> Do we know what precisely this registration affects or is needed for? I'm pretty sure we don't do this for any of the hvsocket stuff we use in the GCS today.

I dont know if the general case requires it, or if doing it allows for some side benefit, besides the docs saying its needed. 
For `hcsshim`, WCOW uses the HCS service id ([1](https://github.com/microsoft/hcsshim/blob/8c246952f89e1e0c7eb9e1dbb97249201b5a13b2/internal/uvm/create_wcow.go#L75), [2](https://github.com/microsoft/hcsshim/blob/8c246952f89e1e0c7eb9e1dbb97249201b5a13b2/internal/gcs/protocol.go#L26-L31)). 
At least for me, thats already registered in my registry.

And for LCOW, we use the [well known Vsock service ID](https://github.com/microsoft/hcsshim/blob/8c246952f89e1e0c7eb9e1dbb97249201b5a13b2/internal/uvm/create_lcow.go#L822)

kevpar:
> > Do we know what precisely this registration affects or is needed for? I'm pretty sure we don't do this for any of the hvsocket stuff we use in the GCS today.
> 
> I dont know if the general case requires it, or if doing it allows for some side benefit, besides the docs saying its needed. For `hcsshim`, WCOW uses the HCS service id ([1](https://github.com/microsoft/hcsshim/blob/8c246952f89e1e0c7eb9e1dbb97249201b5a13b2/internal/uvm/create_wcow.go#L75), [2](https://github.com/microsoft/hcsshim/blob/8c246952f89e1e0c7eb9e1dbb97249201b5a13b2/internal/gcs/protocol.go#L26-L31)). At least for me, thats already registered in my registry.
> 
> And for LCOW, we use the [well known Vsock service ID](https://github.com/microsoft/hcsshim/blob/8c246952f89e1e0c7eb9e1dbb97249201b5a13b2/internal/uvm/create_lcow.go#L822)

Reading the doc from the PR description, I think we likely don't actually need this support. WMI based management is the older model used by VMMS. I suspect this registry entry basically allows the service to be used with a VMMS VM. However with HCS we instead directly configure what services are allowed using [ServiceTable](https://github.com/microsoft/hcsshim/blob/master/internal/hcs/schema2/hv_socket_system_config.go#L21) in the HCS schema, which also gives us much more fine-grained control.

helsaawy:
> Reading the doc from the PR description, I think we likely don't actually need this support. WMI based management is the older model used by VMMS. I suspect this registry entry basically allows the service to be used with a VMMS VM. However with HCS we instead directly configure what services are allowed using [ServiceTable](https://github.com/microsoft/hcsshim/blob/master/internal/hcs/schema2/hv_socket_system_config.go#L21) in the HCS schema, which also gives us much more fine-grained control.

Testing locally, service registration doesn't seem to have effect if using the well-known VM IDs (ie Loopback/Host/...).
I added this mostly for external users that may want it, but if you think thats unnecessary, I'll close this.

dcantah:
My $0.02 at least, I like the helpers. I know we'd been chuggin along due to the pre-registered gcs service ID so we don't need it on our end (at least for our container work), but it seems like a harmless time save for users of the package. 

> I suspect this registry entry basically allows the service to be used with a VMMS VM.

vmms is the only place I really see the key being checked and not just set, so you're probably right.

kevpar:
My preference would be we don't try to support anything VMMS related. It's WMI based (gross) and probably not a rabbit hole we want to go down. HCS all the way :)

helsaawy:
Noted; I'll close this.

PR #239, opened 2022-03-23
Added HV Socket known IDs, Dial, bug fixes

helsaawy:
Added:
* Well-know Hyper-V VMIDs for parents, children, and loopback.
* VSock interop service GUID.
* `Dial()` and `DialContext()` to dial a specific Hyper-V socket at a known address (along with a corresponding `HvsockDialer` struct.

Bug fixes:
* Dial (and Listen) now properly initialize and set properties of their sockets after ConnectEx (and AcceptEx).
* The `socketError` used by `bind` was incorrect, it should be `int32(-1)`, not `uintptr(^0)`

Created a `sockets` package, currently only with syscalls to `Bind`, `ConnectEx` and `GetSockName`, bypassing `syscall/windows` restrictions on the types that cane be used


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/Microsoft/go-winio
 
-go 1.13
+go 1.17
```
kevpar, line 3: I guess even if this affects compat it's okay because we won't necessarily backport it, and 1.17 is the oldest supported Go release?

kevpar: Still would be nice to do things like this as a separate change, though

hvsock.go
```diff
@@ -12,15 +13,73 @@ import (
...
+	// - Physical host: Not supported.
+	HVguidParent = guid.GUID{ // a42e7cda-d03f-480c-9cc2-a4de20abb878
+		Data1: 0xa42e7cda,
+		Data2: 0xd03f,
+		Data3: 0x480c,
+		Data4: [8]uint8{0x9c, 0xc2, 0xa4, 0xde, 0x20, 0xab, 0xb8, 0x78},
+	}
+
+	// HVguidVSockServiceGUIDTemplate is the Service GUID used for the VSOCK protocol
+	hvguidVSockServiceTemplate = guid.GUID{ // 00000000-facb-11e6-bd58-64006a7986d3
```
kevpar, line 78: Not exported?

kevpar: Oh nevermind, we expose a function to get it with the right port.

go.mod
dcantah: Was talking to Hamza about this yesterday. The idea was to get this in as that separate change https://github.com/microsoft/go-winio/pull/230.

> I guess even if this affects compat it's okay because we won't necessarily backport it, and 1.17 is the oldest supported Go release?

Right, there was no plans to backport. We were thinking to start a v0.6.x line of tags.

kevpar:
This seems like a change that should be split across multiple PRs. For instance can we add `sockets` package, and then use it in another PR?

I'd also like to understand better what the motivation is for these changes. What can a user of go-winio _not_ do without these changes?

hvsock.go
```diff
@@ -12,15 +13,73 @@ import (
...
 )
 
-//sys bind(s syscall.Handle, name unsafe.Pointer, namelen int32) (err error) [failretval==socketError] = ws2_32.bind
+const afHvSock = 34 // AF_HYPERV
+
+var (
+	// https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#vmid-wildcards
+
+	// HVguidWildcard is the wildcard VmId for accepting connections from all partitions.
+	HVguidWildcard = guid.GUID{} // 00000000-0000-0000-0000-000000000000
```
kevpar, line 28: I would suggest sticking with the current `Hvsock` prefix used in winio, and Go idioms, and name these like `HvsockGUIDWildcard`, `HvsockGUIDBroadcast`, etc.

helsaawy:
> This seems like a change that should be split across multiple PRs. For instance can we add `sockets` package, and then use it in another PR?
> 
> I'd also like to understand better what the motivation is for these changes. What can a user of go-winio _not_ do without these changes?

Without the `sockets` package, a user wouldn't be able to call `GetPeerName`, `Bind`, or `ConnectEx` on an arbitrary socket type (the `syscall` packages only allow them on types implemented internal to them). We had our own `bind` internally here, but it wasnt exported and required handling raw pointers. 

Those three (`GetPeerName`, `Bind`, and`ConnectEx`) are fairly useless on their own, but they allow use to easily define our own `Dial` for HV sockets.

I could add `sockets` alone, but I think it would look fairly unnecessary, as without Hvsock defining its own socket type, it would appear fairly redundant with golang's socket functions

hvsock.go
kevpar: The last word in the name should still be capitalized (`Wildcard`, `Loopback`, etc.).

hvsock.go
```diff
@@ -12,15 +13,73 @@ import (
...
+	// - The silo host of a silo on a physical host is the physical host.
+	HvsockGUIDSiloHost = guid.GUID{ // 36bd0c5c-7276-4223-88ba-7d03b654c568
+		Data1: 0x36bd0c5c,
+		Data2: 0x7276,
+		Data3: 0x4223,
+		Data4: [8]byte{0x88, 0xba, 0x7d, 0x03, 0xb6, 0x54, 0xc5, 0x68},
+	}
+
+	// HvsockGUIDchildren is the wildcard VmId for accepting connections from the connector's child partitions.
+	HvsockGUIDchildren = guid.GUID{ // 90db8b89-0d35-4f79-8ce9-49ea0ac8b7cd
```
kevpar, line 57: HvsockGUIDChildren

hvsock.go
```diff
@@ -12,15 +13,73 @@ import (
...
 )
 
-//sys bind(s syscall.Handle, name unsafe.Pointer, namelen int32) (err error) [failretval==socketError] = ws2_32.bind
+const afHvSock = 34 // AF_HYPERV
+
+var (
+	// https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#vmid-wildcards
+
+	// HvsockGUIDWildcard is the wildcard VmId for accepting connections from all partitions.
+	HvsockGUIDWildcard = guid.GUID{} // 00000000-0000-0000-0000-000000000000
```
kevpar, line 28: I am actually wondering if we should make these all functions that return a `guid.GUID`, since otherwise the global value could be mutated. What do you think?

helsaawy: valid, I'll update accordingly.

go.mod
dcantah, line 3: Needs a rebase as we swapped to this a bit ago

hvsock.go
```diff
@@ -12,16 +14,87 @@ import (
...
+	return guid.GUID{
+		Data1: 0xffffffff,
+		Data2: 0xffff,
+		Data3: 0xffff,
+		Data4: [8]uint8{0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff},
+	}
+}
+
+// HvsockGUIDLoopback is the Loopback VmId for accepting connections to the same partition as the connector.
+func HvsockGUIDLoopback() guid.GUID { // e0e16197-dd56-4a10-9195-5ee7a155a838
```
dcantah, line 44: Are these not globals so nobody mucks around with them / can't make constant structs?

hvsock.go
```diff
@@ -257,13 +507,19 @@ func (conn *HvsockConn) IsClosed() bool {
...
-		return ErrFileClosed
+		return sockets.ErrSocketClosed
 	}
 
 	err := syscall.Shutdown(conn.sock.handle, how)
 	if err != nil {
+		// If the connection was closed, shutdowns fail with "not connected"
+		if errors.Is(err, windows.WSAENOTCONN) ||
+			errors.Is(err, windows.WSAESHUTDOWN) {
+			err = sockets.ErrSocketClosed
```
dcantah, line 521: I don't like discarding the actual error msg from the OS, in some cases it can be extremely helpful in tracking down what actually happened. I'd see if we can come up with a custom error of our own that preserves this data.

pkg/sockets/sockets.go
```diff
@@ -0,0 +1,175 @@
...
+// BOOL LpfnConnectex(
+//   [in]           SOCKET s,
+//   [in]           const sockaddr *name,
+//   [in]           int namelen,
+//   [in, optional] PVOID lpSendBuffer,
+//   [in]           DWORD dwSendDataLength,
+//   [out]          LPDWORD lpdwBytesSent,
+//   [in]           LPOVERLAPPED lpOverlapped
+// )
+func connectEx(s windows.Handle, name unsafe.Pointer, namelen int32, sendBuf *byte, sendDataLen uint32, bytesSent *uint32, overlapped *windows.Overlapped) (err error) {
```
dcantah, line 165: Can we define this the mksyscall route?

hvsock.go
helsaawy: We could define our own error type/instance for socket closed that wraps it, but I was thinking that most users would not care about the actual error code number, just that the socket was closed. And it makes it easier to check with `errors.Is(err, net.ErrClosed)` or `errors.Is(err, sockets.ErrSocketClosed)`

I have been thinking about moving the error code listings from hcsshim here, so we may add a pkg\errdefs` here that lets us accomplish both...

pkg/sockets/sockets.go
dcantah: I didn't fully get the wsaioctl bits on the comment that's my bad

hvsock.go
```diff
@@ -136,21 +236,54 @@ func (l *HvsockListener) Accept() (_ net.Conn, err error) {
...
+		&addrbuf[0],
+		0, // rxdatalen
+		addrlen,
+		addrlen,
+		&bytes,
+		&c.o)
+	_, err = l.sock.asyncIo(c,
+		nil, // deadlineHandler
+		bytes,
+		err)
```
dcantah, line 256: I really don't know what the best practice for multi line function calls, but this is hard to read for me (it's likely a mix of two multiline functions with no newline in between also haha)

hvsock.go
```diff
@@ -136,21 +236,54 @@ func (l *HvsockListener) Accept() (_ net.Conn, err error) {
...
+	// socket, and is sometimes the same as the local address of the socket that dialed the
+	// address, with the service GUID.Data1 incremented, but othertimes is different.
+	// todo: does the local address matter? is the listener's address or the actual address appropriate?
+
+	// var ra rawHvsockAddr
+	// err = sockets.GetSockName(windows.Handle(sock.handle), &ra)
+	// if err != nil {
+	// 	return nil, conn.opErr("accept", os.NewSyscallError("getsockname", err))
+	// }
+	// conn.local.fromRaw(&ra)
```
dcantah, line 285: We should remove leftover commented out blocks

hvsock.go
```diff
@@ -160,38 +293,160 @@ func (l *HvsockListener) Close() error {
...
 	if err != nil {
-		return nil, err
+		return nil, conn.opErr(op, err)
 	}
 	defer sock.wg.Done()
 	var bytes uint32
-	err = windows.ConnectEx(windows.Handle(sock.handle), sa, nil, 0, &bytes, &c.o)
-	_, err = sock.asyncIo(ctx, c, nil, bytes, err)
+	n := 1 + int(d.Retries)
+	for i := 1; i <= n; i++ {
```
dcantah, line 355: Dumb q.. why start at 1?

helsaawy: for `i` or `n`? I guess `i := 0; i <n` makes more sense, so I can change that

internal/socket/rawaddr.go
```diff
@@ -0,0 +1,41 @@
...
+
+// todo: should these be custom types to store the desired/actual size and addr family?
+
+var (
+	ErrBufferSize     = errors.New("buffer size")
+	ErrInvalidPointer = errors.New("invalid pointer")
+	ErrAddrFamily     = errors.New("address family")
+)
+
+// todo: replace this with generics, along with GetSockName and co.
```
dcantah, line 17: If you plan on taking these up at some point, probably best to `TODO: helsaawy - ...` for todos

internal/socket/rawaddr.go
```diff
@@ -0,0 +1,41 @@
...
+
+// RawSockaddr allows structs to be used with Bind and ConnectEx. The
+// struct must meet the Wind32 sockaddr requirements specified here:
```
dcantah, line 20: win32*

internal/socket/socket.go
```diff
@@ -0,0 +1,174 @@
...
+//sys getsockname(s windows.Handle, name unsafe.Pointer, namelen *int32) (err error) [failretval==socketError] = ws2_32.getsockname
+//sys getpeername(s windows.Handle, name unsafe.Pointer, namelen *int32) (err error) [failretval==socketError] = ws2_32.getpeername
+//sys bind(s windows.Handle, name unsafe.Pointer, namelen int32) (err error) [failretval==socketError] = ws2_32.bind
+
+const socketError = uintptr(^uint32(0))
+
+var ErrSocketClosed = fmt.Errorf("socket closed: %w", net.ErrClosed)
+
+// CloseWriter is a connection that can disable writing to itself.
+type CloseWriter interface {
```
dcantah, line 27: Are these used/returned from anywhere here?

dcantah: This and CloseReader

hvsock.go
```diff
@@ -160,38 +293,160 @@ func (l *HvsockListener) Close() error {
 	return l.sock.Close()
 }
 
-/* Need to finish ConnectEx handling
-func DialHvsock(ctx context.Context, addr *HvsockAddr) (*HvsockConn, error) {
+type HvsockDialer struct {
```
dcantah, line 296: Comment for the exported dialer

internal/socket/socket.go
helsaawy: `ErrSocketClosed` is returned hvsock.go, CloseReader/Writer are leftovers, ill remove

dcantah: Idk why it highlighted this whole block.. just meant CloseWriter and CloseReader 😐 but okay sweet! Resolving

hvsock.go
dcantah: Right, and thanks!

hvsock.go
```diff
@@ -12,16 +14,87 @@ import (
...
+		Data1: 0xa42e7cda,
+		Data2: 0xd03f,
+		Data3: 0x480c,
+		Data4: [8]uint8{0x9c, 0xc2, 0xa4, 0xde, 0x20, 0xab, 0xb8, 0x78},
+	}
+}
+
+// HVguidVSockServiceGUIDTemplate is the Service GUID used for the VSOCK protocol
+func hvsockVsockServiceTemplate() guid.GUID { // 00000000-facb-11e6-bd58-64006a7986d3
+	return guid.GUID{
```
dcantah, line 92: These are out of sync now

hvsock.go
```diff
@@ -65,18 +140,43 @@ func (addr *HvsockAddr) fromRaw(raw *rawHvsockAddr) {
 	addr.ServiceID = raw.ServiceID
 }
 
+// Sockaddr interface allows use with `sockets.Bind()` and `.ConnectEx()`
+func (r *rawHvsockAddr) Sockaddr() (ptr unsafe.Pointer, len int32, err error) {
+	return unsafe.Pointer(r), int32(unsafe.Sizeof(rawHvsockAddr{})), nil
```
dcantah, line 145: Can we do the validations in this function itself instead of have the caller do it after this? Referring to the steps in validateSockAddr:

```go
ptr, n := unsafe.Pointer(r), int32(unsafe.Sizeof(rawHvsockAddr{})
if ptr == nil {
	return nil, 0, fmt.Errorf("pointer is %p: %w", ptr, ErrInvalidPointer)
}
if n < 1 {
	return nil, 0, fmt.Errorf("buffer size %d < 1: %w", n, ErrBufferSize)
}
return ptr, n, nil

Comment (dcantah) [reply hvsock.go:145]: I guess for the buffer size, unless we completely butchered a change and removed every field in rawHvsockAddr we'd never hit that

File: hvsock.go [hunk 18]
```diff
@@ -160,38 +274,159 @@ func (l *HvsockListener) Close() error {
...
 	if err != nil {
-		return nil, err
+		return nil, conn.opErr(op, err)
 	}
 	defer sock.wg.Done()
 	var bytes uint32
-	err = windows.ConnectEx(windows.Handle(sock.handle), sa, nil, 0, &bytes, &c.o)
-	_, err = sock.asyncIo(ctx, c, nil, bytes, err)
+	n := 1 + int(d.Retries)
+	for i := 0; i < n; i++ {
```
dcantah, line 337: Can we do `i <= n`? Wouldn't that get us to where we need without the i < (n-1) below also? We'd run through once and if we failed we'd retry n times. Is the idea if we fail we retry n times or the first attempt is part of these number of retries

hvsock.go
helsaawy: my intention was that since `(*rawHvsockAddr) Sockaddr()` was user defined, and internally we could have a bug in how we define it, `validateSockAddr` is always going to run and prevent segfaults or panics
But, looking at it now, I guess we can use reflection to get the pointer and struct size, like how `json.Unmarshall` does it ....

hvsock.go
```diff
@@ -15,72 +16,85 @@ import (
...
+	"github.com/Microsoft/go-winio/internal/socket"
 	"github.com/Microsoft/go-winio/pkg/guid"
-	"github.com/Microsoft/go-winio/pkg/sockets"
 )
 
 const afHvSock = 34 // AF_HYPERV
 
-var (
-	// https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#vmid-wildcards
+// Well know Service and VM IDs
```
dcantah, line 25: known*

hvsock.go
```diff
@@ -279,6 +274,7 @@ func (l *HvsockListener) Close() error {
 	return l.sock.Close()
 }
 
+// HvsockDialler configures and dials a Hyper-V Socket (ie, [HvsockConn]).
```
dcantah, line 277: HvsockDialer*

hvsock.go
```diff
@@ -293,11 +289,11 @@ type HvsockDialer struct {
 	rt *time.Timer // redial wait timer
 }
 
-func (d *HvsockDialer) Dial(addr *HvsockAddr) (*HvsockConn, error) {
-	return d.DialContext(context.Background(), addr)
+func Dial(ctx context.Context, addr *HvsockAddr) (conn *HvsockConn, err error) {
```
dcantah, line 292: Comment for exported funcs

internal/socket/socket.go
```diff
@@ -0,0 +1,163 @@
...
+// buffer returned by the getsockname syscall. The buffer is allocated to the size specified
+// by `rsa.Sockaddr()`.
+func GetSockName(s windows.Handle, rsa RawSockaddr) error {
+	ptr, l, err := rsa.Sockaddr()
+	if err != nil {
+		return fmt.Errorf("could not find socket size to allocate buffer: %w", err)
+	}
+	if err = validateSockAddr(ptr, l); err != nil {
+		return err
+	}
```
dcantah, line 36: I'd still like for the impls to verify inside

go.mod
dcantah: Going to resolve this as the 1.17 upgrade happened awhile ago already

internal/socket/socket.go
helsaawy: Done, and since we assume the caller is returning a valid pointer, we can also get rid of `FromBytes` and overwrite the memory pointed to ourselves.

hvsock.go
```diff
@@ -206,8 +450,8 @@ func (conn *HvsockConn) Read(b []byte) (int, error) {
 	err = syscall.WSARecv(conn.sock.handle, &buf, 1, &bytes, &flags, &c.o, nil)
 	n, err := conn.sock.asyncIo(c, &conn.sock.readDeadline, bytes, err)
 	if err != nil {
-		if _, ok := err.(syscall.Errno); ok {
-			err = os.NewSyscallError("wsarecv", err)
+		if eno := windows.Errno(0); errors.As(err, &eno) {
```
dcantah, line 453: I'd prefer 
```go
var eno windows.Errno
if error.As(err, &eno) {}
```

dcantah: Same for the others

hvsock.go
```diff
@@ -160,38 +277,165 @@ func (l *HvsockListener) Close() error {
...
-	_, err = sock.asyncIo(ctx, c, nil, bytes, err)
+	for i := uint(0); i <= d.Retries; i++ {
+		err = socket.ConnectEx(
+			windows.Handle(sock.handle),
+			&sa,
+			nil, // sendBuf
+			0,   // sendDataLen
+			&bytes,
+			(*windows.Overlapped)(unsafe.Pointer(&c.o)))
+		_, err = sock.asyncIo(c, nil, bytes, err)
```
dcantah, line 354: I'd love if we can find a way to pass our timeouts to/thread a context through asyncIo. I'm just thinking, and correct my misunderstanding if there is one, if we set some deadline of 100ms lets say, and waiting for the connection exceeds that, we won't return immediately with what's here now. Connection attempts not in a goroutine, and there's no select waiting on a completion or ctx timeout. asyncIo does, but we haven't passed anything for the deadlineHandler and it doesn't accept a context

helsaawy: we wouldn't, no
I tried jerry-rigging a context into a `deadlineHandler` where i launched a go-routine that cancelled the `deadlineHandler` `chan` when the goroutine was cancelled, but that proved to be ugly and sometimes took too long for all the necessary go routines to be called

helsaawy: I have an approach I am working on, but I would rather move that to another PR

dcantah: > I have an approach I am working on, but I would rather move that to another PR

Sure I'm fine with this

dcantah (approved):

anmaxvl (approved):

PR #240, opened 2022-03-23
Added HV Socket tests

helsaawy:
Added tests for core Hyper-V socket functionality, including testing
CloseRead and CloseWrite, as well as checking addresses are appropriate
and timeouts work.

Requires #239 

### DIFF
--- a/hvsock_test.go
+++ b/hvsock_test.go
@@ -18,21 +18,27 @@ import (
 	"github.com/Microsoft/go-winio/pkg/guid"
 )
 
-// TODO: timeouts on listen
-
 const testStr = "test"
 
 func randHvsockAddr() *HvsockAddr {
-	p := uint32(rand.Int31())
+	p := rand.Uint32() //nolint:gosec // used for testing
 	return &HvsockAddr{
 		VMID:      HvsockGUIDLoopback(),
 		ServiceID: VsockServiceID(p),
 	}
 }
 
-func serverListen(u testUtil) (*HvsockListener, *HvsockAddr) {
-	a := randHvsockAddr()
-	l, err := ListenHvsock(a)
+func serverListen(u testUtil) (l *HvsockListener, a *HvsockAddr) {
+	var err error
+	for i := 0; i < 3; i++ {
+		a = randHvsockAddr()
+		l, err = ListenHvsock(a)
+		if errors.Is(err, windows.WSAEADDRINUSE) {
+			u.T.Logf("address collision %v", a)
+			continue
+		}
+		break
+	}
 	u.Must(err, "could not listen")
 	u.T.Cleanup(func() {
 		if l != nil {
@@ -45,21 +51,18 @@ func serverListen(u testUtil) (*HvsockListener, *HvsockAddr) {
 
 func clientServer(u testUtil) (cl, sv *HvsockConn, _ *HvsockAddr) {
 	l, addr := serverListen(u)
-	ch := make(chan struct{})
-	go func() {
-		defer close(ch)
-
+	ch := u.Go(func() error {
 		conn, err := l.Accept()
-		u.Must(err, "listener accept")
+		if err != nil {
+			return fmt.Errorf("listener accept: %w", err)
+		}
 		sv = conn.(*HvsockConn)
-		u.T.Cleanup(func() {
-			if sv != nil {
-				u.Must(sv.Close(), "server close")
-			}
-		})
-		u.Must(l.Close())
+		if err := l.Close(); err != nil {
+			return err
+		}
 		l = nil
-	}()
+		return nil
+	})
 
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
@@ -71,8 +74,12 @@ func clientServer(u testUtil) (cl, sv *HvsockConn, _ *HvsockAddr) {
 		}
 	})
 
-	u.Wait(ch, time.Second)
-	u.Check()
+	u.WaitErr(ch, time.Second)
+	u.T.Cleanup(func() {
+		if sv != nil {
+			u.Must(sv.Close(), "server close")
+		}
+	})
 	return cl, sv, addr
 }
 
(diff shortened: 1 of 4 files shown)


hvsock_test.go
```diff
@@ -0,0 +1,627 @@
...
+	for _, b := range [][]byte{
+		[]byte("hello?"),
+		[]byte("This is a really long string that should be a good example of the really long " +
+			"payloads that may be sent over hvsockets when really long inputs are being used, tautologically. " +
+			"That means that we will have to test with really long input sequences, which means that " +
+			"we need to include really long byte sequences or strings in our testing so that we know that " +
+			"the sockets can deal with really long inputs. Look at this key mashing: " +
+			"sdflhsdfgkjdhskljjsad;kljfasd;lfkjsadl ;fasdjfopiwej09q34iur092\"i4o[piwajfliasdkf-012ior]-" +
+			"01oi3;'lSD<Fplkasdjgoisaefjoiasdlj\"hgfoaisdkf';laksdjdf[poaiseefk-0923i4roi3qwjrf9" +
+			"08sEJKEFOLIsaejf[09saEJFLKSADjf;lkasdjf;kljaslddhgaskghk"),
```
dcantah, line 515: Commit of the year nominee

dcantah: Not to take away from the incredible sequence of characters here, but you can also make a multiline string with backquotes

```go

[]byte(`This is a really long string that should be a good example of the really long 
			payloads that may be sent over hvsockets when really long inputs are being used, tautologically. 
			That means that we will have to test with really long input sequences, which means that 
			we need to include really long byte sequences or strings in our testing so that we know that 
			the sockets can deal with really long inputs. Look at this key mashing: 
			sdflhsdfgkjdhskljjsad;kljfasd;lfkjsadl ;fasdjfopiwej09q34iur092\"i4o[piwajfliasdkf-012ior]-
			01oi3;'lSD<Fplkasdjgoisaefjoiasdlj\"hgfoaisdkf';laksdjdf[poaiseefk-0923i4roi3qwjrf9
			08sEJKEFOLIsaejf[09saEJFLKSADjf;lkasdjf;kljaslddhgaskghk`),
```

hvsock_test.go
```diff
@@ -0,0 +1,627 @@
...
+	defer t.Stop()
+	select {
+	case <-ch:
+	case <-t.C:
+		u.T.Helper()
+		u.T.Fatalf(_msgJoin(msgs, "timed out after %v"), d)
+	}
+}
+
+func _msgJoin(pre []string, s string) string {
```
dcantah, line 625: Any reason for the underscore? Generally not very idiomatic

hvsock_test.go
```diff
@@ -0,0 +1,627 @@
...
+		defer close(ch)
+		c, err := l.Accept()
+		u.Must(err, "listener accept")
+		defer c.Close()
+
+		hv := c.(*HvsockConn)
+		//
+		// test CloseWrite()
+		//
+		_, err = c.Write([]byte(testStr))
```
dcantah, line 298: Additionally verify we wrote the whole payload?

dcantah: Well, the contract for an io.Writer states we need to return an error if we wrote < len(byteSlice), so if we got an error check the n is what I mean.

hvsock_test.go
helsaawy: partially old habit, but also, since `hvsock_test.go` shares the same (testing) namespace for all of `winio`, I wanted to keep this hidden to prevent conflicts
I haven't come up with a good way to do so ....
(In my defense, they use underscores in the [go source code](https://cs.opensource.google/search?q=_%5Ba-z%5D&ss=go%2Fgo), and for auto-generated mkwinsyscall code....)

hvsock_test.go
helsaawy: wait, raw strings include the newlines and I cant escape things (though I doubt thats an issue)

hvsock_test.go
dcantah: TIL, looks like they use it in tests a bit and for syscall definitions they don't wanna export but want to somewhat keep the screaming case for, _BLAH_BLAH.

helsaawy: I removed the leading `_`, hopefully we can isolate the hvsock tests (or testing support) in the future

hvsock_test.go
```diff
@@ -0,0 +1,631 @@
...
+		u.Must(err, "server tx")
+
+		ch <- struct{}{}
+
+		_, err = c.Write([]byte(testStr))
+		u.Must(err, "server tx")
+
+		c.Close()
+	}()
```
dcantah, line 391: Fatal calls in a goroutine just stop the goroutine itself and not the test. Can we have these goroutines send to an error chan inside instead and we can read on the main thread? It looks like the pattern of:

Make chan -> goroutine for listener/dial -> send/close to chan on success might also warrant a new helper for your testing utility, maybe you pass in an anonymous func that makes use of a channel and the channel in question is made in the method and returned. e.g. something like:

```go
func (h *hi) Go(run func(chan error)) chan error {
	ch := make(chan error)
	go func() {
		run(ch)
	}()
	return ch
}


errCh := u.Go(func(ch chan error){
     ch <- errors.New("oh no")
})


err <- errCh

```

helsaawy: I actually toyed around with this a bunch, but in `run`, we would still have to write
```go
errCh := u.Go(func(ch chan error){
    if err := doSomething(...); err != nil {
        ch <- fmt.Errorf("oh no: %w", err)
        return
    }
    moreThings(...)
})
u.Must(<- errCh)
```

and instead `u.Must` and co. give you give you the `if err != nil { return }` for free, so you avoid a lot of boilerplate
The only downside is we assume that you always want to stop a test (`t.FailNow()`) immediately on failure instead of just calling `t.Fail()`, but I haven't see cases where we do the latter.

I have `u.Check()` so that way you can check if a go-routine failed (via `t.Failed()`), and in the cases I do need to synchronize (eg, `TestHvSockCloseReadWriteDial`), I do use a channel as a block.

helsaawy: But also, I think your point is valid and this isnt the most flexible approach--I'm just hesitant to do a redesign in this PR 😅
We should have a more detailed design discussion in the future about testing primitives....

dcantah: I mean the signature could even be:

```go
func (u *util) Go(run func() error ) chan error {
	ch := make(chan error)
	go func() {
		ch <- run()
	}()
	return ch
}

func TestBlah(t *testing.T) {
	errCh := Go(func() error {
		if err := stuff(); err != nil {
			return err
		}
		return nil
	})

	u.Wait(errCh, time.Second*5)
}
```

Now you just write standard error handling for your concurrent bits, and can just pass the channel to your Wait().

kevpar: Are we just recreating [`errgroup.Group`](https://pkg.go.dev/golang.org/x/sync/errgroup#Group) here?

helsaawy: > Are we just recreating [`errgroup.Group`](https://pkg.go.dev/golang.org/x/sync/errgroup#Group) here?

Kind of, but 
 1. we dont always need to wait for the goroutine to finish running; and 
 2.  the main point of the `util` is to avoid having to write:
```go
if err != nil {
    return err
}
// ...
return nil
```
or 
```go
if !errors.Is(err, target) {
    return fmt.Errorf("got error %v, wanted %v", err, target)
}
// ...
return nil
```

rarely here or in our other tests do we do anything besides ending the test when an error (or the wrong one) occurs, so it adds more overhead to constantly have to add return statements everywhere.
Especially since the main test routine will eventually just print error to stdout and then call `FailNow()`

Additionally, `errorgroup.Group` only has a `g.Wait() error` function, so to use it in a select statement (ie, to add a timeout), you need to do pass in a context. And that context's cancellation doesn't propagate to the error group, so you have to use the `ctx.Done()` or the returned context in conjunction with another timeout

anmaxvl (approved):

anmaxvl:
oops, didn't see the last bit of conversation 😄

hvsock_test.go
dcantah: I think my main point is it feels weird to have a model of:

Kill goroutine if error ->  wait in main thread -> channel gets closed by virtue of the `defer close(ch)` so Wait wakes up -> check if it failed after Wait()

Vs.

Send nil or error to chan -> Wait in main thread for error/nil/timeout from goroutine -> t.Fatal if timeout or err != nil

helsaawy: Yes, but it saves typing 😅😅
I guess your suggestion would be cleaner:
```go 
func (u *util) Go(run func() error ) chan error {
	ch := make(chan error)
	go func() {
		ch <- run()
	}()
	return ch
}

func (u testUtil) WaitErr(ch <-chan error, d time.Duration, msgs ...string) {
	t := time.NewTimer(d)
	defer t.Stop()
	select {
	case err := <-ch:
		u.Must(err, msgs...)
	case <-t.C:
		u.T.Helper()
		u.T.Fatalf(msgJoin(msgs, "timed out after %v"), d)
	}
}
```

helsaawy: its just irksome that `test` has a utility function to end a test early, but it cannot be used from a goroutine....

dcantah (approved):

PR #241, opened 2022-03-29
Updating windows build constraints

helsaawy:
#191 added build guard to go-winio, but "pkg/guid" now works on both windows and linux, and is not OS dependent.
This PR is needed to allow using windows-style GUIDs on Linux code that interops with Windows.

Removed build guards from pkg/guid/guid.go.
Build constraints to guid_windows and guid_nonwindows, which redundant with file names, but added for consistency.

Added missing build constraints to other files.

Added Ubuntu to test matrix, along with windows 2022.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,10 @@ on:
 
 jobs:
   test:
-    runs-on: 'windows-2019'
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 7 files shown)


katiewasnothere (approved):
LGTM, I don't see any issue doing this

anmaxvl (approved):

PR #242, opened 2022-04-06
Assign ErrPipeListenerClosed to net.ErrClosed

dcantah:
This error used to just be a `errors.New` with the text exactly matching the "use of closed network connection" that you'd get from the Go stdlib. Now that that error was exported as `net.ErrClosed` in Go 1.16 we should
be able to swap to this safely.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
+        with:
+          go-version: '^1.17.0'
       - run: go test -gcflags=all=-d=checkptr -v ./...
 
   build:
--- a/pipe.go
+++ b/pipe.go
@@ -89,8 +89,7 @@ const (
 
 var (
 	// ErrPipeListenerClosed is returned for pipe operations on listeners that have been closed.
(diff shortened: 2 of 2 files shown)


dcantah:
CI's using Go 1.15.5 and that's the reason for the fail, will look in a little

helsaawy (approved):

dcantah:
@thaJeztah I could only see here where moby was making use of this, and it seems like the usage would continue working. Maybe our comment was out of date 🤷‍♂️ https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/libcontainerd/remote/client_io_windows.go#L84-L86. 

There's these two spots that check the error text but they're both using tcp/http conns, so not really relevant. 
1. https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/integration/internal/requirement/requirement.go#L20
2. https://github.com/moby/moby/blob/7b9275c0da707b030e62c96b679a976f31f929d3/cmd/dockerd/metrics.go#L27

dcantah:
fyi @katiewasnothere @ambarve as I think we'd talked about this before

dcantah:
@katiewasnothere @ambarve Anybody able to give this a look?

ambarve (approved):
LGTM!

PR #243, opened 2022-04-06
Extend GrantVmGroupAccess to support write/execute/all permissions.

anmaxvl:
Add masks for GENERIC_WRITE, GENERIC_EXECUTE, GENERIC_ALL and update
function signatures accordingly.

Add unit tests for file and directory grant access.

Signed-off-by: Maksim An <maksiman@microsoft.com>

### DIFF
--- a/pkg/security/grantvmgroupaccess.go
+++ b/pkg/security/grantvmgroupaccess.go
@@ -37,7 +37,12 @@ type (
 )
 
 const (
-	accessMaskDesiredPermission accessMask = 1 << 31 // GENERIC_READ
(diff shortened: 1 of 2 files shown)


kevpar:
Given how VM-specific this code is, maybe we should just move it under `internal` in hcsshim instead?

anmaxvl:
> Given how VM-specific this code is, maybe we should just move it under `internal` in hcsshim instead?

yeah, sure. I can do that.

anmaxvl:
Port PR: https://github.com/microsoft/hcsshim/pull/1347

PR #245, opened 2022-04-20
Set ETW event name and options for logrus hook

helsaawy:
Added functionality to set the event (task) name and options (such as
event or associated event ID) for ETW events created by the hook based
on the logrus.Entry fields.

This PR supports enhanced logging in the shim.

### DIFF
--- a/pkg/etwlogrus/hook.go
+++ b/pkg/etwlogrus/hook.go
@@ -1,35 +1,73 @@
+//go:build windows
 // +build windows
 
 package etwlogrus
 
 import (
+	"errors"
 	"sort"
 
-	"github.com/Microsoft/go-winio/pkg/etw"
 	"github.com/sirupsen/logrus"
+
+	"github.com/Microsoft/go-winio/pkg/etw"
 )
+
+const DefaultEventName = "LogrusEntry"
+
+var ErrNoProvider = errors.New("no ETW registered provider")
+
+// HookOpt is an option to change the behavior of the Logrus ETW hook
+type HookOpt func(*Hook) error
 
 // Hook is a Logrus hook which logs received events to ETW.
 type Hook struct {
 	provider      *etw.Provider
 	closeProvider bool
+	// allows setting the entry name
+	getName func(*logrus.Entry) string
+	// returns additional options to add to the event
+	getEventsOpts func(*logrus.Entry) []etw.EventOpt
 }
 
 // NewHook registers a new ETW provider and returns a hook to log from it. The
 // provider will be closed when the hook is closed.
-func NewHook(providerName string) (*Hook, error) {
-	provider, err := etw.NewProvider(providerName, nil)
-	if err != nil {
-		return nil, err
-	}
+func NewHook(providerName string, opts ...HookOpt) (*Hook, error) {
+	opts = append(opts, WithNewETWProvider(providerName))
 
-	return &Hook{provider, true}, nil
+	return NewHookFromOpts(opts...)
 }
 
 // NewHookFromProvider creates a new hook based on an existing ETW provider. The
 // provider will not be closed when the hook is closed.
-func NewHookFromProvider(provider *etw.Provider) (*Hook, error) {
-	return &Hook{provider, false}, nil
+func NewHookFromProvider(provider *etw.Provider, opts ...HookOpt) (*Hook, error) {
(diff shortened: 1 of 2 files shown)


pkg/etwlogrus/hook.go
```diff
@@ -1,35 +1,73 @@
...
+	"errors"
 	"sort"
 
-	"github.com/Microsoft/go-winio/pkg/etw"
 	"github.com/sirupsen/logrus"
+
+	"github.com/Microsoft/go-winio/pkg/etw"
 )
 
+const DefaultEventName = "LogrusEntry"
```
kevpar, line 15: Do we have to export this?

pkg/etwlogrus/hook.go
```diff
@@ -1,35 +1,73 @@
...
 
-	return &Hook{provider, true}, nil
+	return NewHookFromOpts(opts...)
 }
 
 // NewHookFromProvider creates a new hook based on an existing ETW provider. The
 // provider will not be closed when the hook is closed.
-func NewHookFromProvider(provider *etw.Provider) (*Hook, error) {
-	return &Hook{provider, false}, nil
+func NewHookFromProvider(provider *etw.Provider, opts ...HookOpt) (*Hook, error) {
```
kevpar, line 42: Suggest we omit this function as users can simply pass `WithExistingETWProvider` themselves.

pkg/etwlogrus/hook.go
```diff
@@ -58,6 +96,19 @@ func (h *Hook) Fire(e *logrus.Entry) error {
...
 	}
 
+	name := DefaultEventName
+	if h.getName != nil {
+		if n := h.getName(e); n != "" {
+			name = n
+		}
+	}
+
+	opts := make([]etw.EventOpt, 0, 3)
```
kevpar, line 106: Why do we initialize capacity to 3?

pkg/etwlogrus/hook.go
helsaawy: I didnt know if an end user would want it for code, otherwise we are setting it without them knowing what it is.

pkg/etwlogrus/hook.go
helsaawy: If we're comfortable with that API break, I'll remove it.

pkg/etwlogrus/hook.go
helsaawy: I wanted to be generous and give users space for two extra options without having incur a realloc + copy

pkg/etwlogrus/hook.go
kevpar: I think it's fine to leave unexported. They can just override it if they want to.

pkg/etwlogrus/hook.go
kevpar: Oh I missed that this was already present. That's fine then.

pkg/etwlogrus/hook.go
kevpar: A comment explaining would be helpful, since otherwise the inclination is to try and understand what is causing us to use 3 elements.

helsaawy:
@kevpar resolved, PTAL

pkg/etwlogrus/hook.go
kevpar: Let's please add a little wording like "just in case the user passed a couple of options, so we don't need to reallocate quite as often" to the end

kevpar (approved):
One comment on wording but otherwise LGTM

pkg/etwlogrus/hook.go
```diff
@@ -1,35 +1,73 @@
...
 
-	"github.com/Microsoft/go-winio/pkg/etw"
 	"github.com/sirupsen/logrus"
+
+	"github.com/Microsoft/go-winio/pkg/etw"
 )
 
+const defaultEventName = "LogrusEntry"
+
+var ErrNoProvider = errors.New("no ETW registered provider")
```
dcantah, line 17: Comment for godoc

dcantah (approved):
LGTM

pkg/etwlogrus/hook.go
dcantah: Just line 17, not sure why it selected multiple..

PR #248, opened 2022-05-23
Update mkwinsyscall and make stand alone tool

helsaawy:
Updated `mkwinsyscall` to latest version, and created a standalone tool that can be called from go-winio and hcsshim.
Used latest version from:
https://github.com/golang/sys/blob/bc2c85ada10aa9b6aa9607e9ac9ad0761b95cf1d/windows/mkwinsyscall/mkwinsyscall.go

Applied changes from both `go-winio` (forced UTF-16, `go-winio` import) and `hcsshim` (UTF-16, `HResult` error return type)

Added (temporary?) flag to disable sorting so that generated syscall file diffs in hcsshim are easier to read.

Added build constraint comment to generated file, in case file names do not have one.

Allowed input files can be specified with a glob pattern. 
This allows `hcsshim/internal/winapi` to avoid listing all the files within it:
`//go:generate go run github.com/Microsoft/go-winio/tools/mkwinsyscall -output zsyscall_windows.go ./*.go`

PR has two commits so that changes can be rebased on newer versions of `golang.org/x/sys/windows/mkwinsyscall`, but it may make more sense to create a dedicated branch on this repo with those commits


syscall.go
```diff
@@ -1,3 +1,3 @@
 package winio
 
-//go:generate go run golang.org/x/sys/windows/mkwinsyscall -output zsyscall_windows.go file.go pipe.go sd.go fileinfo.go privilege.go backup.go hvsock.go
+//go:generate go run github.com/Microsoft/go-winio/tools/mkwinsyscall -output zsyscall_windows.go file.go pipe.go sd.go fileinfo.go privilege.go backup.go hvsock.go
```
dcantah, line 3: Couldn't we use the glob functionality here?

dcantah:
This is great that we're converging on one copy. Have you verified vendoring a local copy of winio with this change to hcsshim has everything work alright?

syscall.go
helsaawy: I did not think of that, I'll switch

helsaawy:
> This is great that we're converging on one copy. Have you verified vendoring a local copy of winio with this change to hcsshim has everything work alright?

None of the tests fail and the diffs dont look too egregious here: https://github.com/microsoft/hcsshim/pull/1409.
Unfortunately, I dont think we have really good testing around our syscalls

tools/mkwinsyscall/mkwinsyscall.go
```diff
@@ -580,26 +573,26 @@ func (f *Fn) HelperCallParamList() string {
...
+	errorVar := f.Rets.ErrorVarName()
 	if errorVar == "" {
 		errorVar = "err"
 	}
-	return fmt.Sprintf(code, errorVar, p.DLLFuncName())
+	return fmt.Sprintf(code, errorVar, f.DLLFuncName())
 }
 
 // IsUTF16 is true, if f is W (utf16) function. It is false
 // for all A (ascii) functions.
```
dcantah, line 592: Should update the comment to talk about the second condition this now checks (utf16 flag)

dcantah (approved):

anmaxvl (approved):

PR #256, opened 2022-08-22
GUID tests to use t.Run

helsaawy:
Using subtests with `t.Run()` for GUID testing.

Added `.String` to GUID `Variant` and `Version` using `golang.org/x/tools/cmd/stringer`.

Added `tools.go` to version `stringer` in go.mod and allow `go generate ./...` to be run without needing a `go get` call.
Based off of [go wiki](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) recommendation.


tools.go
```diff
@@ -0,0 +1,5 @@
+//go:build tools
+
+package winio
+
+import _ "golang.org/x/tools/cmd/stringer"
```
kevpar, line 5: Is the goal here to control what version of `golang.org/x/tools/cmd/stringer` is used in `go generate` via the `go.mod` file?

helsaawy: yup, I have a PR somewhere to add something similar to hcsshim for `github.com/josephspurrier/goversioninfo/cmd/goversioninfo` and `github.com/Microsoft/go-winio/tools/mkwinsyscall`.
But also, I dont know the full details, but if its not in go.mod, then `go run golang.org/x/tools/cmd/stringer` (and therefore `go generate`) sometimes fails for me

kevpar: What failure do you get? I haven't heard of that before.

helsaawy:
```pwsh
> go generate .\pkg\guid\
no required module provides package golang.org/x/tools/cmd/stringer; to add it:
        go get golang.org/x/tools/cmd/stringer
pkg\guid\guid.go:17: running "go": exit status 1
```

Though running `go get` adds it to `go.mod`, but with an indirect flag:
```
golang.org/x/tools v0.1.12 // indirect
```

So I guess we can manually track tools via `go get`, but I went the go tools route following the [Go recommendation](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module).

helsaawy:
rebased onto main to fix merge conflicts

tools.go
kevpar: Ah cool, I wasn't aware of this recommendation. That link would be a good thing to include in your commit description.

anmaxvl (approved):

.github/workflows/ci.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       - name: Run golangci-lint
         uses: golangci/golangci-lint-action@v3
         with:
+          version: v1.48
```
kevpar, line 21: What's the story here?

helsaawy: Opps, need to rebase to remove that

kevpar (approved):
LGTM

PR #260, opened 2022-10-09
tools/mkwinsyscall: remove errnoERROR_IO_PENDING const

thaJeztah:
- related https://github.com/golang/sys/pull/140

From the looks of it, this const is redundant (but perhaps there's something I'm overlooking 😅)

### DIFF
--- a/internal/socket/zsyscall_windows.go
+++ b/internal/socket/zsyscall_windows.go
@@ -15,12 +15,8 @@ var _ unsafe.Pointer
 
 // Do the interface allocations only once for common
 // Errno values.
-const (
-	errnoERROR_IO_PENDING = 997
-)
-
 var (
-	errERROR_IO_PENDING error = syscall.Errno(errnoERROR_IO_PENDING)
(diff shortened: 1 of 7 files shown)


helsaawy (approved):
From what i am able to tell, `errnoERROR_IO_PENDING` was originally added for `WSAEINPROGRESS`, which leads me to believe it was added like the because syscall doesnt define it as an error number:
https://go-review.googlesource.com/c/go/+/28484/4/src/syscall/mksyscall_windows.go

Since `windows` defers to`syscall.Errno`, this shouldnt cause weird type conversion issues either....

helsaawy (changes requested):
you will need to add a `"windowsdot": windowsdot,` line in the `funcMap` on line 852 so that `windowsdot` in the template works

thaJeztah:
Rebased, and fixed;

> you will need to add a `"windowsdot": windowsdot,` line in the `funcMap` on line 852 so that `windowsdot` in the template works


cc @jterry75

PR #261, opened 2022-10-11
pkg/fs: Add RemoveAll function

kevpar:
New RemoveAll function intended to be a replacement for os.RemoveAll usage and be tailored to Windows scenarios.

Most importantly, this function works with []uint16 instead of string. This allows it to properly delete Windows filesystem trees that contain files with invalid UTF-16 paths.


kevpar:
I still want to determine if long path support is needed here. However the code should be reasonable to review as it exists already.

kevpar:
Looks like some overzealous linter errors. May need to tune those down.

pkg/fs/removeall_windows.go
```diff
@@ -0,0 +1,214 @@
...
+
+// We want to use findFirstFileExW since it allows us to avoid the shortname lookup.
+// Package golang.org/x/sys/windows only defines FindFirstFile, so we define our own here.
+// We could use its findNextFileW definition, except that it also has an inaccurate version of Win32finddata[1].
+// Therefore we define our own versions of both, as well as our own findData struct.
+//
+// [1] This inaccurate version is still usable, as the name and alternate name fields are one character short,
+// but the original value is guaranteed to be null terminated, so you just lose the null terminator in the worst case.
+// However, this is still annoying to deal with, as you either must find the null in the array, or treat the full array
+// as a null terminated string.
```
helsaawy, line 19: Can we move this discussion right above `findData`, since it explains why we arent using `windows.Win32finddata`?

pkg/fs/removeall_windows.go
```diff
@@ -0,0 +1,214 @@
...
+// where a string-based API would not, such as when a filesystem name is not valid UTF-16 (e.g. low surrogates without preceding high surrogates).
+//
+// RemoveAll handles some dynamic changes to the directory tree while we are deleting it. For instance,
+// if a directory is not empty, we delete its children first, and then try again. If new items are added to the
+// directory while we do that, we will again recurse and delete them, and keep trying. This matches the behavior of
+// os.RemoveAll.
+// However, we don't attempt to handle every dynamic case, for instance:
+//   - If a file has FILE_ATTRIBUTE_READONLY re-set on it after we clear it, we will fail.
+//   - If an entry is changed from file to directory after we query its attributes, we will fail.
+func RemoveAll(path []uint16) error {
```
helsaawy, line 83: can we add an alias or new type `UTF16Str` for []uint16, so in the future we can do our own validation and add helper functions (like filepath join and co) to work on windows path strings?

pkg/fs/removeall_windows.go
```diff
@@ -0,0 +1,214 @@
...
+	return append(append(parent, '\\'), child...)
+}
+
+func pathErr(op string, path []uint16, err error) error {
+	return &fs.PathError{Op: op, Path: windows.UTF16ToString(path), Err: err}
+}
+
+// terminate takes a []uint16 and returns a null-terminated *uint16.
+func terminate(path []uint16) *uint16 {
+	return &append(path, '\u0000')[0]
```
helsaawy, line 202: should we check that this path isnt already null-terminated?

pkg/fs/removeall_windows.go
```diff
@@ -0,0 +1,214 @@
...
+	for i := range v1 {
+		if v1[i] != v2[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func join(parent, child []uint16) []uint16 {
+	return append(append(parent, '\\'), child...)
```
helsaawy, line 193: this would cause duplicate `\`'s if parent already ends with one?

pkg/fs/removeall_windows.go
```diff
@@ -0,0 +1,214 @@
...
+					}
+					err = findNextFileW(find, &fd)
+					if err == windows.ERROR_NO_MORE_FILES {
+						break
+					} else if err != nil {
+						return pathErr("FindNextFile", path, err)
+					}
+				}
+				return nil
+			}(); err != nil {
```
helsaawy, line 159: Can this be moved out to a separate function?
Ideally we can wrap `findFirstFileExW` and `findNextFileW` into a `walkFiles(f func(findData) error)) error`, since that allows the defer to still work, and I am sure we'll need it again in the future

helsaawy (changes requested):

pkg/fs/removeall_windows.go
kevpar: I wanted it here as a reference for why we define our own syscalls. I can look at what I could put in the other spot as well though.

pkg/fs/removeall_windows.go
kevpar: I considered adding a type but wasn't sure how much utility it would have, or what a good name would be. `UTF16Str` is exactly what we _don't_ want since the whole point of this function is it can work with a name that is invalid UTF-16. :)

kevpar: I can think more about what a good name would be.

pkg/fs/removeall_windows.go
kevpar: I don't think there's any valid case where we would pass a null terminated string here, since the idea is that we only terminate right as we pass it to a syscall. Maybe we could add a check, but it would be the equivalent of a debug assert semantically.

pkg/fs/removeall_windows.go
kevpar: Let me think about if there's a path where we would expect to need to handle that here. Maybe just the case where the user passes in a path that ends in a separator.

pkg/fs/removeall_windows.go
```diff
@@ -0,0 +1,214 @@
...
+//
+// https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-file_attribute_tag_infof
+type fileAttributeTagInfo struct {
+	attributes uint32
+	tag        uint32
+}
+
+var (
+	// Mockable for testing.
+	removeDirectory = windows.RemoveDirectory
```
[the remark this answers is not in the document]
kevpar: I don't see any way to avoid it here, and don't think the cost is too high. We don't just need to log operations done, but actually need to inject a failure case into a Windows API call (look at `TestRemoveAllShouldFailWhenSymlinkDeletionFails`).

pkg/fs/removeall_windows.go
kevpar: I'm not sure here. I think it can be easier to grok the overall flow when it's kept together in a single function, and didn't want to introduce more abstractions like a walk function when the existing code seems fairly straightforward. However the complexity did grow over time, so maybe it would make sense.

pkg/fs/removeall_windows.go
helsaawy: Maybe `OSString`, to steal Rust's thunder?

pkg/fs/removeall_windows.go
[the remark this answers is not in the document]
helsaawy: but rather than allowing the `removeDirectory` function to be modified, we can do something like

```go
type testHook func(p []uint16)

var removeDirectoryTestHook testHook

func removeDirectory(path []uint16) error {
  if removeDirectoryTestHook != nil {
    removeDirectoryTestHook(path)
  }
  windows.RemoveDirectory
}
```

And then `TestRemoveAllShouldFailWhenSymlinkDeletionFails` becomes
```go
var linkDeleteAttempted bool
removeDirectoryTestHook = func(path *uint16) {
  if _, name := filepath.Split(windows.UTF16PtrToString(path)); !linkDeleteAttempted && name == "link" {
    linkDeleteAttempted = true
  }
}

if err := RemoveAll(rootU16); err != nil {
  // ...
}

if linkDeleteAttempted {
  t.Fatal("link deleted")
}
```

TBBle:
For the record, [Go 1.21 will obviate the need for this this](https://github.com/golang/go/issues/32334#issuecomment-1550003371) by using WTF-8 to carry filenames on Windows, ensuring round-trip-safe `string` representation of the full range of valid Windows filenames.

kevpar:
> For the record, [Go 1.21 will obviate the need for this this](https://github.com/golang/go/issues/32334#issuecomment-1550003371) by using WTF-8 to carry filenames on Windows, ensuring round-trip-safe `string` representation of the full range of valid Windows filenames.

Yeah, I think it probably makes sense to just wait for go1.21 at this point. Glad that change made it in. :)

PR #262, opened 2022-10-20
WriteEventString writes a single ETW event from the provider.

thaJeztah:
This function allows writing a single, non-structured (textual) ETW event.

- relates to https://github.com/moby/moby/pull/44289

In moby/moby, we were maintaining a local implementation for our etwlogs driver (originally implemented in https://github.com/moby/moby/pull/19689).

I'm in the process of replacing that implementation with the code from go-winio (https://github.com/moby/moby/pull/44289) so that we can have a single "canonical" implementation; most bits were already supported by go-winio, except for this one.

⚠️ I realise this is not the "right" approach; this PR is mostly a "straight" port from what we had. Like other functions in this package;

- it _probably_ needs the 32/64 bit wrappers (are those still needed?)
- it should be generated using `mkwinsyscall` -I started to write the generate comment like [the existing ones](https://github.com/microsoft/go-winio/blob/376b2013fee695d54f9b21db81716426a8a736d9/pkg/etw/syscall.go#L5-L15), but as I don't have a Windows machine at hand, had no way to verify them 😅)

So, I thought; let me open this as a "proposal" PR to see if adding this function is acceptable, and _hoping_ someone could help moving it over the finish line (write and test the generator comments).


thaJeztah:
@kevpar @helsaawy ptal 🤗 

I'm also wondering of (some of) these would make sense to be upstreamed to golang.org/x/sys/windows (if so, I'm happy to draft up a PR for that).

helsaawy:
@thaJeztah 
I implemented what I think is the best of both worlds [here](https://github.com/helsaawy/go-winio/commit/2dae0398f0e2ca322e2470e8add57a7fc888fc9a) based on the [docs saying this function should not be used](https://learn.microsoft.com/en-us/windows/win32/api/evntprov/nf-evntprov-eventwritestring).

It logs at level 0 without a task name (just like your code), but the events will have the message in a field called "Messge" instead of just containing the raw string. I do not think this will be a problem (though let me know if otherwise), and keeps with the code aligned with [TraceLogging](https://learn.microsoft.com/en-us/windows/win32/tracelogging/trace-logging-portal).

thaJeztah:
> based on the [docs saying this function should not be used](https://learn.microsoft.com/en-us/windows/win32/api/evntprov/nf-evntprov-eventwritestring).

Interesting. I'll be honest; I don't know a lot about this area. Curious why that specific API was picked at the time.

In general, I think (in moby) we should probably look if we can change these logs to be more structured, but that might be a breaking change, so my first goal was to keep the existing approach, but unify implementations (where possible).

> I dont know what the correct way to collaborate on a PR is, maybe I need to open a PR against your branch?

My PR was a very quick one, that (effectively) just copies the code we have in Moby.

Feel free to open a separate PR with your changes 👍 

> It logs at level 0 without a task name (just like your code), but the events will have the message in a field called "Messge" instead of just containing the raw string. I do not think this will be a problem (though let me know if otherwise), and keeps with the code aligned with [TraceLogging](https://learn.microsoft.com/en-us/windows/win32/tracelogging/trace-logging-portal).

Sounds good! (as mentioned) I'm not very familiar with this, so could use some quick 👀 from others to confirm that change would be ok; let's do so on your PR (once you have time to open one).

Thank you!

helsaawy:
Created #263, if you cant to close this in favor of that

thaJeztah:
Closing in favour of https://github.com/microsoft/go-winio/pull/262 👍

PR #263, opened 2022-10-24
Add WriteEventString function.

helsaawy:
This is an implementation of #262.

Adds a `(*Provider).WriteEventString` to write a single string to ETW
Since the intent is to add an intermediary function for Moby to call while they transition to this library, the function is marked as deprecated, to ward against future use.

### DIFF
--- a/pkg/etw/provider.go
+++ b/pkg/etw/provider.go
@@ -220,6 +220,21 @@ func (provider *Provider) IsEnabledForLevelAndKeywords(level Level, keywords uin
 	return true
 }
 
+// WriteEventString writes a single ETW event from the provider that contains the string
+// as its data with the field name set to "Message".
+//
+// It is a wrapper around [Provider.WriteEvent], and meant to replace directly calling the
+// [ETW WriteEventString] syscall, which should not be used.
+//
+// Deprecated: Use [Provider.WriteEvent] instead.
+//
+// [ETW WriteEventString]: https://learn.microsoft.com/en-us/windows/win32/api/evntprov/nf-evntprov-eventwritestring#remarks
+func (provider *Provider) WriteEventString(message string) error {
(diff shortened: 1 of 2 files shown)


helsaawy:
@thaJeztah @kevpar

kevpar:
This seems very application-specific to moby. Rather than add an already-deprecated function here, wouldn't it be easier for moby to just define their own function and use that while transitioning to structured logging?

```golang
func WriteEventString(p *etw.Provider, m string) error {
	return p.WriteEvent("StringMessage", []etw.EventOpt{etw.WithLevel(0)}, []etw.FieldOpt{etw.StringField("message", m)})
}
```

helsaawy:
> This seems very application-specific to moby. Rather than add an already-deprecated function here, wouldn't it be easier for moby to just define their own function and use that while transitioning to structured logging?
> 
> ```go
> func WriteEventString(p *etw.Provider, m string) error {
> 	return p.WriteEvent("StringMessage", []etw.EventOpt{etw.WithLevel(0)}, []etw.FieldOpt{etw.StringField("message", m)})
> }
> ```

as in: 
```go
type Provider struct {
    etw.Provider
}

func (p *Provider) WriteString(string) { /* ... */ }
```

helsaawy:
@thaJeztah 
Would you be able to do something like below in moby as an interim until you fully transition?
```go
type Provide struct {
  *etw.Provider
}

func (p *etw.Provider) WriteEventString(m string) error {
	return p.WriteEvent("StringMessage", []etw.EventOpt{etw.WithLevel(0)}, []etw.FieldOpt{etw.StringField("message", m)})
}
```

helsaawy:
Okay, I'll close this then.
In the meantime; enjoy  kubecon :)

PR #266, opened 2022-11-09
go.mod: remove golang.org/x/tools dependency

thaJeztah:
This removes golang.org/x/tools/cmd/stringer as a direct dependency of this module, instead using "go install" to install the required tool as part of `go generate`. I opted for pinning to a fixed version for reproducibility, but we could switch to `@latest` if there's no concerns for getting newer versions.

With this patch:

    which stringer
    # (no output)

    go generate ./pkg/guid
    which stringer
    /Users/thajeztah/go/bin/stringer


thaJeztah:
@kevpar @helsaawy ptal 🤗

helsaawy:
I don't think I understand the need for this. 
I agree with removing logrus as a dependency, but for stringer, I would prefer keeping the tools file, since it follows the [recommended approach to version dependencies](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module).

If trimming the dependencies for upstream projects is the goal, starting in [go1.17](https://go.dev/ref/mod#graph-pruning), `stringer` should not be considered a dependency by modules that import winio.

Also, using `go run` doesn't install `stringer` by putting it `$(go env GOPATH)\bin`, but instead builds and runs it out of `go env GOCACHE`, so that also keeps the path clean as well.

jterry75:
This is way old and we just replaced it. So closing.

PR #274, opened 2023-02-04
Add some basic bind filter functions

gabriel-samfira:
This change adds the ability to mount a a single folder or a volume inside another folder, using the bind filter API.

While the API allows mounting multiple sources inside a single mount point, acting as an overlay, we disable this functionality in the ```ApplyFileBinding()``` function. We could add another ```flags``` parameter to this function to allow using any of the existing flags, making it more generic. Let me know if this is preferred.

Tests will be added shortly.

Signed-off-by: Gabriel Adrian Samfira <gsamfira@cloudbasesolutions.com>


TBBle:
Cross-tagging from https://github.com/microsoft/go-winio/pull/187#issuecomment-1419296718: I suggest that once this lands, hcsshim be updated to use this for `wclayer mount` and `wclayer unmount` replacing the Volume Mount Point APIs that #187 was deduplicating, and hence allowing #187 to be sunset as "No current users": The lowest-maintenance-cost code is the code never committed. ^_^

gabriel-samfira:
> I realise the current layout of this repo is... adventurous... but should this API be somewhere under `pkg/` rather than in the root?

I am extremely malleable in regard to this. Any suggestions or preferences are fine with me.

gabriel-samfira:
I added some tests. It would be nice to add some tests with mount points, but I am not sure if pulling in ```computestorage``` dependencies (I think it requires additional windows features?) in order to format a VHD and make it mountable, is desirable.

TBBle:
Thinking more about my earlier thought regarding hcsshim, I just remembered that there's probably [already code _in_ hcsshim that uses Bind Filter](https://github.com/microsoft/hcsshim/blob/420d7fa852e08c17575a1df467497a1929829ed6/internal/jobobject/jobobject.go#L508-L518) (and [a syscall wrapper for the one call they use](https://github.com/microsoft/hcsshim/blob/420d7fa852e08c17575a1df467497a1929829ed6/internal/winapi/bindflt.go), so converting that to use this API would be a good validation of the API, including questions about generalisation.

gabriel-samfira:
> Thinking more about my earlier thought regarding hcsshim, I just remembered that there's [already code _in_ hcsshim that uses Bind Filter](https://github.com/microsoft/hcsshim/blob/420d7fa852e08c17575a1df467497a1929829ed6/internal/jobobject/jobobject.go#L508-L518) (and [a syscall wrapper for the one call they use](https://github.com/microsoft/hcsshim/blob/420d7fa852e08c17575a1df467497a1929829ed6/internal/winapi/bindflt.go)), so converting that to use this API would be a good validation of the API, including questions about generalisation.

Yup. It's in use in hcsshim. I think we can replace *some* things in hcsshim with this, but the ```BfGetMappings``` seems to only be available in ltsc2022. The tests seem to fail on 2019. Initially, this will be used in ```containerd``` to allow us to add support for ```buildkitd```. Come to think of it, we may need to put the ```containerd``` windows snapshotter behind a feature flag or a check that the OS version is supported. We'll circle back to that later, though.

pkg/bindfilter/bind_filter.go
```diff
@@ -0,0 +1,334 @@
...
+	"unicode/utf16"
+	"unsafe"
+
+	"golang.org/x/sys/windows"
+)
+
+//go:generate go run github.com/Microsoft/go-winio/tools/mkwinsyscall -output zsyscall_windows.go ./bind_filter.go
+//sys bfSetupFilter(jobHandle windows.Handle, flags uint32, virtRootPath *uint16, virtTargetPath *uint16, virtExceptions **uint16, virtExceptionPathCount uint32) (hr error) = bindfltapi.BfSetupFilter?
+//sys bfRemoveMapping(jobHandle windows.Handle, virtRootPath *uint16)  (hr error) = bindfltapi.BfRemoveMapping?
+//sys bfGetMappings(flags uint32, jobHandle windows.Handle, virtRootPath *uint16, sid *windows.SID, bufferSize *uint32, outBuffer uintptr)  (hr error) = bindfltapi.BfGetMappings?
```
helsaawy, line 24: for UTF-16 string params, you can just write:
```//sys bfSetupFilter(jobHandle windows.Handle, flags uint32, virtRootPath string, virtTargetPath string, virtExceptions **uint16, virtExceptionPathCount uint32) (hr error) = bindfltapi.BfSetupFilter?```
and the generated code will convert to `*uint16` automatically. I dont think it supports `[]string though`

File: pkg/bindfilter/bind_filter.go [hunk 2]
```diff
@@ -0,0 +1,334 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	flags := BINDFLT_GET_MAPPINGS_FLAG_VOLUME
+	// allocate a large buffer for results
+	var outBuffSize uint32 = 256 * 1024
+	buf := make([]byte, outBuffSize)
+
+	if err := bfGetMappings(flags, 0, rootPtr, nil, &outBuffSize, uintptr(unsafe.Pointer(&buf[0]))); err != nil {
```
helsaawy, line 140: you should be able to specify `//sys .... outBuffer *byte` and the code will convert to `uintptr` (unless im missing something)

pkg/bindfilter/bind_filter.go
```diff
@@ -0,0 +1,334 @@
...
+		return nil, fmt.Errorf("invalid buffer returned")
+	}
+
+	result := buf[:outBuffSize]
+
+	// The first 12 bytes are the three uint32 fields in getMappingsResponseHeader{}
+	headerBuffer := result[:12]
+	// The alternative to using unsafe and casting it to the above defined structures, is to manually
+	// parse the fields. Not too terrible, but not sure it'd worth the trouble.
+	header := *(*getMappingsResponseHeader)(unsafe.Pointer(&headerBuffer[0]))
```
helsaawy, line 154: This is our preferred approach (we do it a lot in this codebase), so no real justification is needed 😅

gabriel-samfira:
> Come to think of it, we may need to put the `containerd` windows snapshotter behind a feature flag or a check that the OS version is supported. We'll circle back to that later, though.

There is no way to list mappings on ltsc2019. We can however blindly try to remove a mapping and ignore the error if it's ```The parameter is incorrect```. That way we can support this starting with RS5.

pkg/bindfilter/bind_filter.go
```diff
@@ -0,0 +1,322 @@
...
+		nil,
+		0,
+	); err != nil {
+		return fmt.Errorf("failed to bind target %q to root %q: %w", source, root, err)
+	}
+	return nil
+}
+
+// RemoveFileBinding removes a mount from the root path.
+func RemoveFileBinding(root string) error {
```
gabriel-samfira, line 101: Might as well rename ```bfRemoveMapping``` to ```RemoveFileBinding``` and remove this function. What do you think?

helsaawy: dealers choice, but if there really isnt any extra setup needed, then that makes the most sense

gabriel-samfira: ahh wait. Spoke too soon. The ```bfRemoveMapping()``` function has the job handle param. Let's leave this as is.

pkg/bindfilter/bind_filter_test.go
```diff
@@ -0,0 +1,211 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestApplyFileBinding(t *testing.T) {
```
helsaawy, line 15: It just occured to me, for testing we sometimes pull the OS version from the registry and use that to skip tests: https://github.com/containerd/containerd/blob/97480afdac09c947d48f5e3a134db86c78f4bfa6/pkg/os/os_windows_test.go#L40-L55

It may be cleaner than relying on skipping tests in the ci file

gabriel-samfira: We'd need to pull in hcsshim as a dependency for the osversion package, or duplicate the values in ```hcsshim/osversion/windowsbuilds.go```, here. Which option would you prefer? :smile:

helsaawy: The later: you can porbably just compare against `17763` instead of copying in all the build numbers
(we probably should move the OS versions from hcsshim to here...)

gabriel-samfira: Done. PTAL.

helsaawy (approved):
LGTM, though I would prefer a review from @msscotb before merging this

pkg/bindfilter/bind_filter.go
```diff
@@ -0,0 +1,322 @@
...
+func getFileHandle(pth string) (syscall.Handle, error) {
+	info, err := os.Lstat(pth)
+	if err != nil {
+		return 0, fmt.Errorf("accessing file: %w", err)
+	}
+	p, err := syscall.UTF16PtrFromString(pth)
+	if err != nil {
+		return 0, err
+	}
+	attrs := uint32(syscall.FILE_FLAG_BACKUP_SEMANTICS)
```
msscotb, line 313: I suspect neither a bind mapping src or target can be name surrogate reparse point like a symlink or junction since bindflt would've resolved the links when setting up its mappings. Need to think about what to do here. Possibly just use syscall.CreateFile directly like https://github.com/containerd/cri/blob/4cc8ca2c07450dc9e477f1486733a2ae67dfd1ba/pkg/os/os_windows.go#L39

pkg/bindfilter/bind_filter.go
```diff
@@ -0,0 +1,322 @@
...
+//sys bfSetupFilter(jobHandle windows.Handle, flags uint32, virtRootPath string, virtTargetPath string, virtExceptions **uint16, virtExceptionPathCount uint32) (hr error) = bindfltapi.BfSetupFilter?
+//sys bfRemoveMapping(jobHandle windows.Handle, virtRootPath string)  (hr error) = bindfltapi.BfRemoveMapping?
+//sys bfGetMappings(flags uint32, jobHandle windows.Handle, virtRootPath *uint16, sid *windows.SID, bufferSize *uint32, outBuffer *byte)  (hr error) = bindfltapi.BfGetMappings?
+
+// BfSetupFilter flags. See:
+// https://github.com/microsoft/BuildXL/blob/a6dce509f0d4f774255e5fbfb75fa6d5290ed163/Public/Src/Utilities/Native/Processes/Windows/NativeContainerUtilities.cs#L193-L240
+//
+//nolint:revive // var-naming: ALL_CAPS
+const (
+	BINDFLT_FLAG_READ_ONLY_MAPPING uint32 = 0x00000001
```
msscotb, line 30: Is there a reason to define all these flags if only BINDFLT_FLAG_NO_MULTIPLE_TARGETS and BINDFLT_FLAG_READ_ONLY_MAPPING are used?

msscotb (changes requested):

pkg/bindfilter/bind_filter.go
gabriel-samfira: The initial scope of this change was to allow us to bind mount a volume or a folder with the option to have that mount as read-only. However, if during review it was decided that we want to make these functions a bit more generic and allow setting the flags when calling the function, the defined flags would already be there.

I've removed the flags. If we need them later, we can always add them back.

pkg/bindfilter/bind_filter.go
gabriel-samfira: Indeed. When applying the bind filter, symlinks ger resolved and the bind mapping ends up pointing to the actual source of the symlink. This means we can just use ```windows.CreateFile()``` without needing to care about reparse points.

Adding a test for that as well.

msscotb (approved):
lgtm

PR #275, opened 2023-02-06
Add `fs.ResolvePath` to resolve symbolic links

helsaawy:
`filepath.EvalSymlinks` does not work well on Windows, and can loop indefinitely in certain situations and error out.

Add `ResolvePath`, which uses the Win32 `GetFinalPathNameByHandle` to resolve the final path of a file by first opening it.
Add `"internal/stringbuffer".WString` struct to maintain a pool of `[]uint16` buffers for use with interacting with the Win32APIs
Add `"internal/fs"` pkg with Win32 `CreateFile`, `GetFinalPathNameByHandle`, and associated flags and access masks

Core implementation is based off of: https://github.com/containerd/containerd/pull/5411


kevpar:
Can we carry over the tests from https://github.com/containerd/containerd/blob/main/pkg/os/os_windows_test.go?

helsaawy:
> Can we carry over the tests from https://github.com/containerd/containerd/blob/main/pkg/os/os_windows_test.go?

I wanted to avoid that since that would `hcsshim` as a dependency for this package, which could cause weird cyclic import issues.
The alternative would be add `computestorage` bindings here, but I would rather save that for a separate PR

helsaawy:
> > Can we carry over the tests from https://github.com/containerd/containerd/blob/main/pkg/os/os_windows_test.go?
> 
> I wanted to avoid that since that would `hcsshim` as a dependency for this package, which could cause weird cyclic import issues. The alternative would be add `computestorage` bindings here, but I would rather save that for a separate PR

@kevpar 
I added tests and needed APIs to this PR: #276

pkg/fs/resolve.go
```diff
@@ -0,0 +1,108 @@
...
+	"errors"
+	"os"
+	"strings"
+
+	"golang.org/x/sys/windows"
+
+	"github.com/Microsoft/go-winio/internal/fs"
+)
+
+// ResolvePath follows symbolic links and returns the resolved path to the file or
```
kevpar, line 15: It does more than resolve symlinks, it also handles mount points, etc.

internal/fs/security.go
```diff
@@ -0,0 +1,12 @@
+package fs
+
+// https://learn.microsoft.com/en-us/windows/win32/api/winnt/ne-winnt-security_impersonation_level
+type SecurityImpersonationLevel int32 // C++ default enums underlying type is `int`, which is Go `int32`
```
kevpar, line 4: It's really a C type, not C++, if you want to be specific here.

internal/fs/security.go
```diff
@@ -0,0 +1,12 @@
...
+
+// Impersonation levels
+const (
+	SecurityAnonymous      = 0
```
kevpar, line 8: Should these be typed constants?

internal/fs/fs.go
```diff
@@ -0,0 +1,251 @@
...
+//	GA     Generic All
+//	Resvd  Reserved
+//	AS     Access Security System
+//
+// https://learn.microsoft.com/en-us/windows/win32/secauthz/access-mask
+//
+// https://learn.microsoft.com/en-us/windows/win32/secauthz/generic-access-rights
+//
+// https://learn.microsoft.com/en-us/windows/win32/fileio/file-access-rights-constants
+type AccessMask uint32
```
kevpar, line 40: I think it's better to use types in golang.org/x/sys/windows when they are available. e.g. https://pkg.go.dev/golang.org/x/sys/windows#ACCESS_MASK

internal/fs/fs.go
```diff
@@ -0,0 +1,251 @@
...
+	// Not actually any.
+	//
+	// For CreateFile: "query certain metadata such as file, directory, or device attributes without accessing that file or device"
+	// https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew#parameters
+	FILE_ANY_ACCESS AccessMask = 0
+
+	// Specific Object Access
+	// from ntioapi.h
+
+	FILE_READ_DATA      AccessMask = (0x0001) // file & pipe
```
kevpar, line 53: Same with constants, if it's in golang.org/x/sys/windows, let's use those instead.

helsaawy: the `sys/windows` constants are not typed, so we lose the ability to type-check and catch bugs that way
and the general constants were all copied in from the corresponding headers, since it was the same amount of work to copy in the ones we needed vs all of them

kevpar: Hmm, the general philosophy has been to carry as little as we can here, and rely on `golang.org/x/sys/windows` for the bulk of Windows support. I'd like to not get in the habit of re-creating functionality that lives in `windows` as well.

Can we get away without using typed constants here? Alternately, can we just add the minimum set of typed constants here that we need? And then we could try to get `windows` to type theirs (though that might be tough, as it is a breaking change).

helsaawy: I doubt they would, since the majority of that code seems auto-generated.
I just copied-pasted the constants here from the relevant header files, so it was trivial to add of them, but ill whittle the number down to just what we use here.

helsaawy: We could do away with the types all together, but I am hesitant to do so, since the function signatures tend to be pretty crowded (eg, `CreateFile(*uint16, uint32, uint32, *windows.SecurityAttributes, uint32, uint32, wiindows.Handle)`) and having types prevents mixing up parameters

kevpar (approved):
The resolve logic looks fine to me. It would be good to have someone else specifically look at the string buffer parts since I didn't focus on that.

internal/fs/fs.go
kevpar: In the interest of not blocking the PR I think this is okay as is. I am interested in longer term how we can push things upstream to stdlib and x/sys/windows rather than accumulating more bits specific to this repo.

msscotb (approved):
lgtm

PR #279, opened 2023-03-08
Add osversion from hcsshim

helsaawy:
Move `osversion` package to `winio` from `hcsshim`, so it can be imported separately.
Additionally, dedicated types can be created for version components, and packaged version number can be removed from `Version` struct without breaking APIs that currently rely on `"hcssim/osversion"`.

[RtlGetVersion](https://learn.microsoft.com/en-us/windows/win32/devnotes/rtlgetversion) returns the windows OS version regardless of manifest, so that is preferred over [GetVersion](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversion).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,8 +4,8 @@ go 1.17
 
 require (
 	github.com/sirupsen/logrus v1.9.0
-	golang.org/x/sys v0.5.0
-	golang.org/x/tools v0.6.0
+	golang.org/x/sys v0.6.0
+	golang.org/x/tools v0.7.0
 )
 
-require golang.org/x/mod v0.8.0 // indirect
+require golang.org/x/mod v0.9.0 // indirect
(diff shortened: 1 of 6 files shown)


tianon:
I use it a _lot_ as the best reference I can find for build numbers and how they correspond to the "public" numbers like 1809 / "LTSC 2019" (I think https://learn.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle is a more canonical location for that, but this Go file is way easier to read for this slice of information).

I've considered using it programmatically several times, but for all my use cases, the fact that it's just a fixed list of constants has been a blocker since it requires manually updating my code any time the list of constants changes (the use case I reconsider frequently is something like converting a particular `os.version` value into a more human-friendly string like "Windows Server 2019" or something).

PR #283, opened 2023-04-08
update to go1.18, drop go1.18, and replace deprecated syscall.Syscall<number>

thaJeztah:
### tools/mkwinsyscall: replace deprecated funcs with SyscallN (drops go1.17)

The "syscall.Syscall<number>" functions were deprecated in go1.18 in favor
of the "syscall.SyscallN" function, which does not need the "nargs" argument,
and does not need the list of arguments to be padded with zeros.
https://github.com/golang/go/blob/go1.18/src/syscall/dll_windows.go#L27-L45

Now that go1.17 reached EOL and is no longer maintained, we can update the
code to use the new SyscallN function. This patch updates the mkwinsyscall
utility to generate code using the new SyscallN function, and removes the
utilities that are now redundant.


internal/socket/zsyscall_windows.go
```diff
@@ -48,23 +48,23 @@ var (
 )
 
 func bind(s windows.Handle, name unsafe.Pointer, namelen int32) (err error) {
-	r1, _, e1 := syscall.Syscall(procbind.Addr(), 3, uintptr(s), uintptr(name), uintptr(namelen))
+	r1, _, e1 := syscall.SyscallN(procbind.Addr(), uintptr(s), uintptr(name), uintptr(namelen))
```
thaJeztah, line 51: :warning: I should add that I patched these manually (don't think I can use `go generate` unless doing so on a Windows machine).

Not 100% sure if CI here verifies these files (if not, perhaps someone could double-check them and run `go generate` on a Windows machine)

thaJeztah: > Not 100% sure if CI here verifies these files (if not, perhaps someone could double-check them and run `go generate` on a Windows machine)

looks like it does, and CI looks happy; https://github.com/thaJeztah/go-winio/actions/runs/4645933496/jobs/8222049635

thaJeztah:
This should be ready for review /cc @kevpar 😄 

(I should probably point out that containerd 1.6 still tests against go1.17 as "old version", in case this version from main ends up in containerd at some point; https://github.com/containerd/containerd/blob/v1.6.20/.github/workflows/ci.yml#L236)

PR #286, opened 2023-04-26
Allow ETW to use thread's activity ID

helsaawy:
This commit is contains several related changes that ultimately allow ETW events to set their activity ID to the current thread's value. By setting the thread's activity ID, win32 API calls from that thread will be able to access it and either continue using it or set it as the related activity ID, improving tracing on Windows.

Update mkwisyscall to import arbitrary other packages, so generated functions can types defined elsewhere.

Switch ETW functions to use `guid.GUID` instead of `windows.GUID`.

Add convenience `IsEmpty` function for GUIDs.

Add `EventActivityIdControl` function to manipulate the thread's activity ID, and related `*ThreadActivityID` functions to set the thread activity ID to a non-empty value.

Update `writeEventRaw` to only pass (related) activity ID to `eventWriteTransfer` if they are non-empty.

This is not a breaking change, since related activity ID will remain the same, and the activity ID will only be changed when not specified if `InitializeThreadActivityID` or `SetThreadActivityID` are called before-hand.

### DIFF
--- a/pkg/etw/activityid.go
+++ b/pkg/etw/activityid.go
@@ -1,0 +1,74 @@
+//go:build windows
+
+package etw
+
+import "github.com/Microsoft/go-winio/pkg/guid"
+
+type eventActivityIDControlCode uint32
+
+//nolint:unused // all values listed here for completeness.
+const (
+	// Sets the ActivityId parameter to the value of the current thread's activity ID.
+	getEventActivityID eventActivityIDControlCode = iota + 1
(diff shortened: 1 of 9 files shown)


kevpar:
What is the motivation behind supporting thread activity ID? Our previous thinking has been this doesn't give much value given the nature of goroutines, and it makes more sense (and is more idiomatic Go) to just carry activity ID around in a context.

Issue #281, opened 2023-03-29
The win32PipeListener.Close() hangs if nobody calls Accept()


thallgren:
The `win32PipeListener.Close()` function attempts to write to an unbuffered channel that has no reader unless someone calls the `Accept()` function. As a result, the call to `Close()` hangs indefinitely. The stack trace from the goroutine that is hanging can look like this:
```
goroutine 84 [chan receive, 5 minutes]:
github.com/Microsoft/go-winio.(*win32PipeListener).Close(0xc0002cdb30)
	github.com/Microsoft/go-winio@v0.6.0/pipe.go:513 +0x85
net/http.(*onceCloseListener).close(...)
	net/http/server.go:3501
sync.(*Once).doSlow(0xc0009a7a50?, 0x1e0d56?)
	sync/once.go:74 +0xc2
sync.(*Once).Do(...)
	sync/once.go:65
net/http.(*onceCloseListener).Close(0xc0005e4bd0?)
	net/http/server.go:3497 +0x4a
net/http.(*Server).closeListenersLocked(0x0?)
	net/http/server.go:2850 +0xad
net/http.(*Server).Shutdown(0xc0002020f0, {0x2690400, 0xc000376500})
	net/http/server.go:2776 +0xa5
```

PR #291, opened 2023-07-07
pipe: add server backlog for concurrent Accept()

jeffhostetler:
Teach `pipe.go:ListenPipe()` to create multiple instances of the server pipe in the kernel so that client connections are less likely to receive a `windows.ERROR_PIPE_BUSY` error. This is conceptually similar to the `backlog` argument of the Unix `listen(2)` function.

The current `listenerRoutine()` function works sequentially and in response to calls to `Accept()`, such that there will only be at most one unbound server pipe present at any time.  Even if the server application calls `Accept()` concurrrently from a pool of application threads, `listenerRoutine()` will process them sequentially.

In this model and because there is only one `listenerRoutine()` instance, there is an interval of time where there are no available unbound/free server pipes.  When `ConnectNamedPipe()` returns `listenerRoutine()` sends the new pipe handle over a channel to the caller of `Accept()`.  Application code then has an opportunity to dispatch/process it and then call `Accept()` again.  This causes `listenerRoutine()` to create a new unbound serer pipe and wait for the next connection.  Anytime during this interval, a client will get a pipe busy error.

Code in `DialPipe()` hides this from GOLANG callers because it includes a busy retry loop.  However, clients written in other languages without this assistance are likely to see it and deal with it.

This change introduces an "accept queue" using a buffered channel and splits `listenerRoutine()` into a pool of listener worker threads.  Each worker creates a new unbound pipe and waits for a client connection.  The NPFS and kernel handle connection delivery to a random listener worker.  The resulting connected pipe is delivered back to the caller `Accept()` as before.

A `PipeConfig.QueueSize` variable controls the number of listener worker threads and the maximum number of unbound/free pipes server pipes that will be present at any given time.  Note that a listener worker will normally have an unbound/free pipe except during that same delivery interval.  Having multiple active workers gives us extra capacity to handle rapidly arriving connections.

The application is encouraged to call `Accept()` from a pool of application workers.  The size of the application pool should be the same or larger than the queue size to take full advantage of the listener queue.

To preserve backwards compatibility, a queue size of 0 or 1 will behave as before.

Also for backwards compatibility, listener workers are required to wait for an `Accept()` call so that the worker has a return channel to send the connected pipe and error code.  This implies that the number of unbound pipes will be the smaller of the queue size and the application pool size.

Finally, a Mutex was added to `l.Close()` to ensure that concurrent threads do not simultaneously try to shutdown the pipe.


jeffhostetler:
I observed this problem while trying to send data from `git.exe` to a GOLANG server.  The code calling `CreateFile()` saw the busy error (or some other error) and didn't expect to need to spin.

I did a little (incomplete) search of the issue backlog and found a few that it might be related:
* https://github.com/microsoft/go-winio/issues/281
* https://github.com/microsoft/go-winio/issues/67#issuecomment-383254610

jeffhostetler:
I added a test at the bottom of `pipe_test.go` that runs various geometries and shows the observed OK-vs-busy rate, but I wasn't sure if/how/when we wanted to throw an error.  On my laptop, the legacy cases get busy errors about 33% of the time.  With a moderate queue size, we don't get busy signals -- however they are still theoretically possible, I just didn't see any on my limited tests, so I hesitated asserting it.

Suggestions welcomed.  Thanks!

PR #293, opened 2023-07-21
Isolate tools dependencies in tools package

helsaawy:
Use `./tools` to prevent callers of `go-winio` from needing to add tools to their dependencies.

Similar to https://github.com/microsoft/hcsshim/pull/1840

Also update dependencies


katiewasnothere (approved):
lgtm

PR #294, opened 2023-07-21
Update tests; run fuzzing

helsaawy:
Run fuzzing tests in CI.

Use race detector when running tests.

Add missing `t.Helper()` calls.

Update test helpers in `pkg/bindfilter` to use `RtlGetNtVersionNumbers` instead of reading registry, and skip tests if not running as admin.

### DIFF
--- a/pkg/bindfilter/bind_filter_test.go
+++ b/pkg/bindfilter/bind_filter_test.go
@@ -8,14 +8,15 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"testing"
 
-	"golang.org/x/sys/windows/registry"
+	"golang.org/x/sys/windows"
 )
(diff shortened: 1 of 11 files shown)


pkg/bindfilter/bind_filter_test.go
```diff
@@ -307,3 +280,23 @@ func TestGetBindMappingsSymlinks(t *testing.T) {
...
+		tb.Skip("requires elevated privileges")
+	}
+}
+
+const RS5 = 17763
+
+//todo: also check that `bindfltapi.dll` exists
+
+// require current build to be >= build
+func requireBuild(tb testing.TB, build uint32) {
```
katiewasnothere, line 296: can't we use the test/pkg/require package instead?

katiewasnothere (approved):

PR #295, opened 2023-07-24
Switch from sycall to windows

helsaawy:
Where ever possible, use `golang.org/x/sys/windows` instead of `syscall` (which has been deprecated since go1.11).

Using `windows.LocalFree` requires using `unsafe.Pointer`, which ensures that the Go garbage collector does not try to free memory pre-maturely if it was previously declared as a pointer.

Since `syscall.Handle` is part of API for `vhd` package, it was left unchanged.

For security descriptor functions, switch to using `windows.SECURITY_DESCRIPTOR` to avoid unnecessary byte manipulation and panics due to missing input validation and error checking.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -11,7 +11,6 @@ import (
 	"net"
 	"os"
 	"runtime"
-	"syscall"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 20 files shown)


pipe.go
```diff
@@ -280,16 +306,20 @@ func makeServerPipeHandle(path string, sd []byte, c *PipeConfig, first bool) (sy
...
 	}
-	defer localFree(ntPath.Buffer)
+	defer windows.LocalFree(windows.Handle(ntPath.Buffer)) //nolint:errcheck
 	oa.ObjectName = &ntPath
 	oa.Attributes = windows.OBJ_CASE_INSENSITIVE
 
 	// The security descriptor is only needed for the first pipe.
 	if first {
 		if sd != nil {
+			//todo: does `sdb` need to be allocated on the heap, or can go allocate it?
```
katiewasnothere, line 316: Did you figure this todo out?

katiewasnothere (approved):

pipe.go
helsaawy: nope, i wasn't able to track it down in the docs or code
i figured we can dig into it in the future, when we clean up the code paths here (and move most of the code to dedicated `pkg/pipe` dir?)

msscotb (approved):

PR #299, opened 2023-08-09
sd.go: fix calculation of security descriptor length in SddlToSecurit…

dblohm7:
…yDescriptor

unsafe.Sizeof(windows.SECURITY_DESCRIPTOR{}) is the minimum length of the SD, not the actual length. Use the actual length for computing the length of the slice.

This path also removes getSecurityDescriptorLength, which is no longer used.


helsaawy (commented):
Linter is failing for an unrelated reason (but that was just fixed: #300), if you can rebase

dblohm7:
@microsoft-github-policy-service agree company="Tailscale Inc."

helsaawy (approved):

PR #305, opened 2023-11-17
Fix/enforce free win32file deadlinehandler

jshen14:
Free up dynamically allocated channel/timer to avoid memory leak


jshen14:

Issue #311, opened 2024-01-12
`FileBasicInfo` alignment is incorrect, leading to `winio.SetFileBasicInfo` returning "Invalid access to memory location."


dagood:
Code that calls `SetFileBasicInfo(f *os.File, bi *FileBasicInfo) error` can in some cases get an error like this:

```
SetFileInformationByHandle \\?\C:\Windows\SystemTemp\hcs2001018820\Files.$wcidirs$: Invalid access to memory location.
```

This is `ERROR_NOACCESS`, which can indicate a `STATUS_DATATYPE_MISALIGNMENT` error.

The reason is that winio's `FileBasicInfo` uses `windows.Filetime`, causing the struct alignment to be wrong. I can't directly prove the cause for sure (it's not trivial to repro with specific functions), but I opened https://github.com/golang/go/issues/65069 to get some help with it, and I don't think there's any doubt.

Here's `FileBasicInfo` for reference:

https://github.com/microsoft/go-winio/blob/bc421d9108ade8ae2f76c8ef3e90cd4ee690e2c0/fileinfo.go#L14-L19

```go
type Filetime struct {
	LowDateTime  uint32
	HighDateTime uint32
}
```

The biggest field of `FileBasicInfo` is `uint32`, so the struct as a whole is aligned on 32 bits.

But the input to `SetFileInformationByHandle` needs to be 64-bit aligned to be compatible with Windows' struct:

> [FileBasicInfo](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_file_basic_information) defines the time properties using `LARGE_INTEGER`, which is a `uint64`.

```
typedef struct _FILE_BASIC_INFORMATION {
  LARGE_INTEGER CreationTime;
  LARGE_INTEGER LastAccessTime;
  LARGE_INTEGER LastWriteTime;
  LARGE_INTEGER ChangeTime;
  ULONG         FileAttributes;
} FILE_BASIC_INFORMATION, *PFILE_BASIC_INFORMATION;
```

---

A workaround is to define a new struct with the right alignment and copy the data over before sending it to the syscall:

```go
func SetFileBasicInfo(f *os.File, bi *FileBasicInfo) error {
	merge := func(t windows.Filetime) uint64 {
		return *(*uint64)(unsafe.Pointer(&t))
	}
	biAligned := struct {
		CreationTime, LastAccessTime, LastWriteTime, ChangeTime uint64
		FileAttributes                                          uint32
[...]

PR #312, opened 2024-01-13
fileinfo: internally fix FileBasicInfo memory alignment

dagood:
* For https://github.com/microsoft/go-winio/issues/311

This PR fixes memory alignment for the structs passed to both `GetFileInformationByHandleEx` and `SetFileInformationByHandle`.

Includes a direct test for memory alignment. (Also makes sure it has the right amount of padding, since it's convenient to check this in the test too.)

### DIFF
--- a/fileinfo_test.go
+++ b/fileinfo_test.go
@@ -6,6 +6,7 @@ package winio
 import (
 	"os"
 	"testing"
+	"unsafe"
 
 	"golang.org/x/sys/windows"
 )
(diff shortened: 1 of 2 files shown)


fileinfo_test.go
```diff
@@ -131,3 +132,49 @@ func TestGetFileStandardInfo_Directory(t *testing.T) {
...
 }
+
+// TestFileInfoStructAlignment checks that the alignment of Go fileinfo structs
+// match what is expected by the Windows API.
+func TestFileInfoStructAlignment(t *testing.T) {
+	//nolint:revive // SNAKE_CASE is not idiomatic in Go, but aligned with Win32 API.
+	const (
+		exampleLARGE_INTEGER uint64 = 0
+		exampleULONGLONG     uint64 = 0
+	)
```
helsaawy, line 143: since `LARGE_INTEGER` is a union with `LONGLONG`, can we replace these two with:
```go
const align64 = unsafe.Alignof(uint64(0))
```

helsaawy (approved):
some nits, but overall LGTM

fileinfo_test.go
dagood: The idea here was to match the type name in the Windows docs with a Go type that represents it to make it easier to compare/review. Less time spent translating. Writing out the `unsafe.Alignof` here rather than later doesn't interfere with that idea though, I'll try that.

PR #314, opened 2024-02-09
Rename module to github.com/microsoft/go-winio

bouk:
The Microsoft GitHub organization uses a lowercase 'm', which other go modules from Microsoft also use. This module using an uppercase M can cause problems in certain build systems when using macOS which uses a case-insensitive file system by default: https://github.com/NixOS/nixpkgs/issues/273998

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/Microsoft/go-winio
+module github.com/microsoft/go-winio
 
 go 1.17
 


bouk:
@helsaawy I was wondering if you could take a look at this?

helsaawy:
Duplicate of PR https://github.com/microsoft/go-winio/pull/159

PR #315, opened 2024-04-01
Update go1.21 and CI

helsaawy:
Update go.mod to go1.21 to match hcsshim.

Use `atomic.Bool` stdlib instead of including our own.

Include `tools\mkwinsyscall` updates from #283 to switch to `syscallN`.
Note: removed `// TODO` about `print`/`ln`, since the latter adds spaces between args when printing, which is undesired.

Update testing code to fix `unchecked-type-assertion` lint.

Remove deprecated fields 
Update CI to:
 - run build and lint steps on windows-2022 instead of windows-2019, similar to our hcsshim CI
 - use latest version of golangci-lint
   - remove deprecated fields in `.golangci.yml` 
 - skip race detector on windows 2019 due to gcc/mingw-related errors


hvsock_test.go
```diff
@@ -123,10 +130,22 @@ func TestHvSockAddresses(t *testing.T) {
...
 	cl, sv, addr := clientServer(u)
 
-	sra := (sv.RemoteAddr()).(*HvsockAddr)
-	sla := (sv.LocalAddr()).(*HvsockAddr)
-	cra := (cl.RemoteAddr()).(*HvsockAddr)
-	cla := (cl.LocalAddr()).(*HvsockAddr)
+	sra, ok := (sv.RemoteAddr()).(*HvsockAddr)
+	if !ok {
+		t.Fatalf("expected type %T; got %T", new(HvsockAddr), sv.RemoteAddr())
+	}
```
kevpar, line 136: You can make this a lot cleaner by defining a function:

```go
func MustBeType[T any](t *testing.T, v any) T {
	v2, ok := v.(T)
	if !ok {
		t.Fatalf("expected type %T; got %T", *new(T), v)
	}
	return v2
}
```

and using it like:

```go
sra := MustBeType[*HvsockAddr](t, sv.RemoteAddr())
```

kevpar:
This PR feels like it has a several unrelated changes, unless there's some interdepenency between them I'm not aware of. If not, it would be best to do these things as multiple PRs in the future. It's easier to review, and easier if we need to track down a specific change later on.

For now though, I'm okay getting this in (pending the other comment I left).

msscotb:
Looks ok but please do split unrelated changes in the future.

msscotb (approved):

helsaawy:
> This PR feels like it has a several unrelated changes, unless there's some interdepenency between them I'm not aware of. If not, it would be best to do these things as multiple PRs in the future. It's easier to review, and easier if we need to track down a specific change later on.
> 
> For now though, I'm okay getting this in (pending the other comment I left).

The switch to go1.21 caused lint errors for the `syscall.N` and the race detector/gcc failure (and some weirdness with golangci-lint v1.53, iirc)
Upgrading golangci-lint introduced warning about the deprecated `.golangci.yml` fields and the `unchecked-type-assertion` linter (added in [v1.55](https://github.com/golangci/golangci-lint/commit/eeb2088101abab41a852390f9c54eddc21c48c85), via [revive v1.3.4](https://github.com/mgechev/revive/commit/95acb880a1b545bdf74dc393a1b80a6df1176cda))

wwqgtxx:
Maybe we don't need to `Update go.mod to go1.21`. 

From the content of this PR, we only need go1.19 to meet all the required language features. 

At the same time, this also makes some users who still need to be compatible with older versions of Golang unable to update from `v0.6.1` to `v0.6.2` 

(for example, it is well known that Golang1.21 has abandoned compatibility with Windows7/8, but there are still many client programs that need to run normally on outdated systems, so they need to continue to use Golang1.20)

Issue #257, opened 2022-08-22
CloseWrite() is non-portable


nicks:
We currently use winio's named pipe implementation to create http duplex sockets.

We use CloseWrite() to close the server side of the socket.

This works ok if the client is also using winio. But if the client uses a different named pipe implemenation, CloseWrite() does not behave correctly. (in particular, NodeJS/libuv's implementation)

My (very naive) understanding is the winio uses a 0-length message to signal the end of the socket. 
https://github.com/microsoft/go-winio/blob/d68e55cd0b80e6e3cac42e945f1cff6ddb257090/pipe.go#L161
which other libraries do not support.

Can you all advise on if there's something we should be doing differently here?

Thank you!

---

Detailed repro steps:

Here's how this breaks things in application-level code: https://github.com/apocas/dockerode/issues/534

Here's a more detailed repro case:

server side:

```
// main.go

package main

import (
	"log"
	"fmt"
	"github.com/Microsoft/go-winio"
)

type CloseWriter interface {
	CloseWrite() error
}

func main() {
	pipe, err := winio.ListenPipe("//./pipe/test-socket", &winio.PipeConfig{
		MessageMode:        true,
		InputBufferSize:    65536,
		OutputBufferSize:   65536,
	})
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("Accepting connections on //./pipe/test-socket")
	for {

		conn, err := pipe.Accept()
		if err != nil {
			log.Fatal(err)
	  }
		fmt.Println("opened socket")
  	go func() {

			conn.Write([]byte("HTTP/1.1 101 Web Socket Protocol Handshake\r\n" +
				"Upgrade: WebSocket\r\n" +
				"Connection: Upgrade\r\n" +
				"\r\n"));
			conn.Write([]byte("closing socket\r\n"))
			conn.(CloseWriter).CloseWrite()
			fmt.Println("closed socket")
		}()
	}
}
```

client side:

```
// index.js
const http = require('http');
const options = {
  headers: {
    'Connection': 'Upgrade',
    'Upgrade': 'tcp'
  },
  socketPath: '//./pipe/test-socket'
};

const req = http.request(options);
req.on('upgrade', (res, socket, upgradeHead) => {
    console.log('connected to socket');
    socket.allowHalfOpen = false;
    socket.on('data', (data) => {
        console.log('socket received data:', data.toString());
    });
    socket.on('end', () => {
        console.log('socket closed');
    });
});
req.end();
```

[...]

Issue #85, opened 2018-07-16
Race condition in `makeConnectedServerPipe` causes `win32PipeListener.Close()` deadlock


djsweet:
On [line 273](https://github.com/Microsoft/go-winio/blob/67921128fb397dd80339870d2193d6b1e6856fd4/pipe.go#L273) in `pipe.go`, `makeConnectedServerPipe` signals to the `connectPipe` goroutine that the listener has been closed by closing the underlying pipe it is trying to connect. However, if the `connectPipe` call succeeds before the call to `p.Close()`, then `err` may also be one of `cERROR_PIPE_CONNECTED` or `cERROR_NO_DATA`. This causes `makeConnectedServerPipe` to effectively swallow the control signal sent over `l.closeCh`, preventing `listenerRoutine` from accepting it, causing `win32PipeListener.Close()` to deadlock without ever closing the listener.

I have been able to reproduce this behavior while developing #84, but only by running `go test` in a loop until the deadlock occurs.

PR #324, opened 2024-12-12
Fix race when closing pipe

mat007:
Hi!

This PR fixes a deadlock that we face every once in a while.

Thanks!

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -574,9 +574,10 @@ func (l *win32PipeListener) Accept() (net.Conn, error) {
 
 func (l *win32PipeListener) Close() error {
 	select {
-	case l.closeCh <- 1:
-		<-l.doneCh
 	case <-l.doneCh:
+	default:
+		close(l.closeCh)
+		<-l.doneCh
 	}
 	return nil
 }
--- a/pipe_test.go
+++ b/pipe_test.go
@@ -646,3 +646,43 @@ func TestListenConnectRace(t *testing.T) {
 		wg.Wait()
 	}
 }
+
+func TestCloseRace(t *testing.T) {
+	for i := 0; i < 200 && !t.Failed(); i++ {
+		l, err := ListenPipe(testPipeName, &PipeConfig{MessageMode: true})
+		if err != nil {
+			t.Fatal(err)
+		}
+		go func() {
+			for {
+				c, err := l.Accept()
+				if err != nil {
+					return
+				}
+				b, err := io.ReadAll(c)
+				if err != nil {
+					t.Error(err)
+					return
+				}
+				_, _ = c.Write(b)
+				_ = c.Close()
+			}
+		}()
+
+		c, err := DialPipe(testPipeName, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if _, err = c.Write([]byte("hello")); err != nil {
+			t.Fatal(err)
+		}
+		if err := c.(CloseWriter).CloseWrite(); err != nil {
+			t.Fatal(err)
+		}
+		if _, err := io.ReadAll(c); err != nil {
+			t.Fatal(err)
+		}
+		_ = c.Close()
+		_ = l.Close()
+	}
+}


thaJeztah:
@mat007 I think this PR fixes https://github.com/microsoft/go-winio/issues/257, correct? If so, can you add a `fixes xxx` to the PR description (not the commit message, as that tends to get noisy 😂)

mat007:
> I think this PR fixes #257, correct?

I don’t know if it fixes that one, I was investigating something else. I’m really not sure it’s related. 🤔

kevpar:
@mat007 can you explain what the deadlock is that you're seeing?

mat007:
> @mat007 can you explain what the deadlock is that you're seeing?

Sure!

The added test without the fix never terminates. Breaking under a debugger we see that the goroutine that called `Close` is blocked on https://github.com/microsoft/go-winio/blob/bdc6c112805ce13d7b235dc6257f258d958f5471/pipe.go#L578 while the `ListenPipe` one is waiting on https://github.com/microsoft/go-winio/blob/bdc6c112805ce13d7b235dc6257f258d958f5471/pipe.go#L462

The problem is that we have 2 readers for `l.closeCh`, with the other one being https://github.com/microsoft/go-winio/blob/bdc6c112805ce13d7b235dc6257f258d958f5471/pipe.go#L447 but we’re signalling it by writing a value to the channel, which can only be read by one reader.
This PR fixes this by closing the channel instead of writing to it.

mat007:
@kevpar what do you think of my explanation? Do you need more info?
Thanks!

kevpar:
> @kevpar what do you think of my explanation? Do you need more info? Thanks!

Sorry, I just got back from holiday. I will take a look today and see if we can get this in!

kevpar:
I've been doing some testing/thinking on this. Overall I think it looks good, just have a few pieces of feedback.

- Let's change `closeCh` and `doneCh` to type `chan struct{}`, since we don't actually care about the value being sent.
- In the `<-l.closeCh` case of `(*win32PipeListener).Close`, I think it's best to still do a `<-l.doneCh` before returning. This will ensure that if there are multiple concurrent calls to `Close`, the later calls will not return until `doneCh` is closed.
- In `TestCloseRace`, can we increase the number of iterations, perhaps to say, 1000? In my testing I sometimes found that it took until iteration 150+ before deadlocking, and we don't want to accidentally get a false negative from this test case if a deadlock is reintroduced.
- Also in `TestCloseRace`, I would suggset adding a `log.Print` statement that prints the iteration number that is completed at the end of the loop. This could help tell when we are deadlocking from test logs if we hit a problem in the future.

Thanks!

kevpar:
Also, looks like there are some CI failures.

thaJeztah:
Had a quick peek; linting errors look unrelated to the PR;

```
Warning: redefines-builtin-id: redefinition of the built-in function max (revive)
Warning: redefines-builtin-id: redefinition of the built-in function max (revive)
Error: printf: non-constant format string in call to (testing.TB).Fatalf (govet)
Error: directive `//nolint:errorlint` is unused for linter "errorlint" (nolintlint)
```

The go-generate one is more interesting, as there's a panic;

```
go run github.com/Microsoft/go-winio/tools/mkwinsyscall -output zsyscall_windows.go socket.go
  go run github.com/Microsoft/go-winio/tools/mkwinsyscall -output zsyscall_windows.go ./bind_filter.go
  go run golang.org/x/tools/cmd/stringer -type=Level -trimprefix=Level
  go: downloading golang.org/x/tools v0.11.0
  go: downloading golang.org/x/mod v0.12.0
  panic: runtime error: invalid memory address or nil pointer dereference [recovered]
  	panic: runtime error: invalid memory address or nil pointer dereference
  [signal 0xc0000005 code=0x0 addr=0x0 pc=0xbff0af]
```



I see https://github.com/microsoft/go-winio/commit/070c828abb873da9e71c7247740253b50f7cf049 changed CI to not used fixed versions of Go, and `oldstable` moved to go1.22; I wonder if the panic is related to that as well;

```
Run actions/setup-go@v5
  with:
    go-version: oldstable
    cache: false
    check-latest: false
    token: ***
  env:
    GO_VERSION: oldstable
    GOTESTSUM_VERSION: latest
    GOLANGCILINT_VERSION: latest
Setup go version spec oldstable
oldstable version resolved as 1.22.10
Found in cache @ C:\hostedtoolcache\windows\go\1.22.10\x64
Added go to the path
Successfully set up Go version oldstable
go version go1.22.10 windows/amd64
```

thaJeztah:
Opened a PR for those linting issues;

- https://github.com/microsoft/go-winio/pull/325


(not sure about the panic, because I'm not on Windows, so maybe it was just a incident; we'll see on the other PR)

mat007:
> I've been doing some testing/thinking on this. Overall I think it looks good, just have a few pieces of feedback.

Thanks! I’ll look into addressing these.

mat007:
@kevpar can you please take another look?
I haven’t tried to fix the lint and go generate failures as they don’t appear to be linked to the changes at hand, and @thaJeztah has opened a separate pull request (thanks @thaJeztah!).

kevpar:
> @kevpar can you please take another look? I haven’t tried to fix the lint and go generate failures as they don’t appear to be linked to the changes at hand, and @thaJeztah has opened a separate pull request (thanks @thaJeztah!).

Changes look good, thanks! One thing that just occurred to me -- there is a chance of a panic if two goroutines try to close a listener at the same time (such that they both try to close `closeCh`). Should we wrap this in a `sync.Once` or something to prevent this?

mat007:
> One thing that just occurred to me -- there is a chance of a panic if two goroutines try to close a listener at the same time (such that they both try to close `closeCh`). Should we wrap this in a `sync.Once` or something to prevent this?

Good catch! This makes the implementation of Close much simpler as well.

kevpar (approved):
LGTM. Thanks for the change :)

kiashok (approved):

mat007:
@kevpar here is a PR to fix the panic with go generate: https://github.com/microsoft/go-winio/pull/326

sorindumitru:
@kevpar We've also been running into this issue quite often as well, mostly in CI (feels like a quarter of the builds fail at least once because of this). Would it be possible to get a release out with the 3 changes from @mat007 (this one, https://github.com/microsoft/go-winio/pull/325 and https://github.com/microsoft/go-winio/pull/326).

mat007:
@kevpar @helsaawy @kiashok what can I do to help move this fix forward?

sorindumitru:
@kevpar @helsaawy @kiashok, pinging again here. We'd really like to have a new release with this fix merged in, it's causing our CI to break very frequently. Are there any plans to look into this PR some more? Or maybe more importantly, is the library still maintained?

sorindumitru:
@helsaawy, @ambarve I see that the linting issues were fixed in https://github.com/microsoft/go-winio/pull/334, maybe all that these needs now is to rebase it on top of main?

mat007:
I’ve rebased the PR, can someone approve the workflow runs?
@kevpar @helsaawy @kiashok @ambarve ?

thaJeztah:
This one was rebased, but had a quick chat with @mat007 and probably the other PR (https://github.com/microsoft/go-winio/pull/369) would indeed fix the main issue. We can still consider keeping this one open to see if there's other parts from this PR that could be useful as a follow-up, but those are less urgent.


From https://github.com/microsoft/go-winio/pull/384#issuecomment-5805064191

Asked my AI to have a quick peek, and looks like the other PR is a smaller, more targeted change to just fix the deadlock (so may be enough for the main issue);

> Both PRs fix the deadlock from #85, but take different approaches.
>
> #369 is the smaller, targeted fix: once `makeConnectedServerPipe` consumes the close signal, it makes that authoritative and returns `ErrPipeListenerClosed` regardless of the result from the raced `ConnectNamedPipe`. That is sufficient to fix the reported deadlock.
>
> #324/#384 changes the underlying close mechanism instead: it closes `closeCh` rather than sending a single value, making it a broadcast signal that cannot be consumed/lost, and uses `sync.Once` to make concurrent `Close` calls safe.
>
> So #369 looks sufficient if we want the minimal fix for #85. #324/#384 is the broader fix if we also want to address the more general issue that `closeCh` represents permanent listener shutdown but is currently implemented as a consumable signal.

jterry75:
@thaJeztah - I think I'd prefer we take this one.

jterry75 (approved):

thaJeztah:
Either way is fine 😅

(I didn't dig deep into the problem myself, so take that with a good pinch of salt)

PR #325, opened 2025-01-10
fix linting issues

thaJeztah:
- relates to https://github.com/microsoft/go-winio/pull/324#issuecomment-2584758696

PR #326, opened 2025-01-30
Update dependencies to fix go generate crash

mat007:
I picked 1.22.10 because I’ve seen this in [another recent PR run](https://github.com/microsoft/go-winio/actions/runs/12718783459/job/35459627702?pr=325):
```
Setup go version spec oldstable
oldstable version resolved as 1.22.10
```
but I’m happy to try and keep it at 1.12 if needed.

Here is what I mainly did to fix the crash:
```
GOTOOLCHAIN=go1.22.10 go get golang.org/x/tools/cmd/stringer
GOTOOLCHAIN=go1.22.10 go mod tidy
```


kiashok (dismissed):
LGTM, thanks!

go.mod
```diff
@@ -1,11 +1,16 @@
 module github.com/Microsoft/go-winio
 
-go 1.21
+go 1.22.0
+
+toolchain go1.22.10
```
thaJeztah, line 5: You can probably remove this one if it automatically added it (go modules tends to randomly add these, but adding also means that users get either down- or upgraded when running commands, so I try to avoid adding these.

go.mod
```diff
@@ -1,11 +1,16 @@
 module github.com/Microsoft/go-winio
 
-go 1.21
+go 1.22.0
```
thaJeztah, line 3: Did the updated modules enforce updating to `go1.22.0`, or does it accept staying on `go1.21` ?

mat007: Yes, it needs go 1.22:
```
$ GOTOOLCHAIN=go1.21.0 go get golang.org/x/tools/cmd/stringer
go: golang.org/x/tools/cmd/stringer: golang.org/x/tools@v0.29.0 requires go >= 1.22.0 (running go 1.21.0; GOTOOLCHAIN=go1.21.0)
```

go.mod
mat007: Yes, I’ll remove the toolchain.

mat007:
@kiashok can you trigger the CI again?

kiashok:
> All right, so this effectively fixes the «CI / Go Generate» workflow. Do you folks want to merge this PR as is, or should we cherry-pick this to #325 to fix all workflows in a single PR?

@mat007 #325 might not fix all the CI issues being hit here right? The interger overflow ones especially - do you want me to look into the integer overflow lint errors? Sorry that a lot of this has fallen on you.

mat007:
> @mat007 #325 might not fix all the CI issues being hit here right? The interger overflow ones especially - do you want me to look into the integer overflow lint errors? Apologies that a lot of this has fallen on you.

Thanks @kiashok ! I can look at the link errors in that other PR, but I need some guidance on how best to approach it (see https://github.com/microsoft/go-winio/pull/325#issuecomment-2626585993)

But also, do we want to merge here, or cherry-pick the commits to #325? I really don’t mind either way. 😄

mat007:
@kiashok sorry for pinging you again.
Is there any chance we can merge this?

PR #335, opened 2025-10-31
fix dangerous uintptr conversion

dangodangodango:
According to: https://pkg.go.dev/unsafe#Pointer `Unsafe.Pointer` rule 4:

> For the compiler to recognize this pattern, the conversion must appear in the argument list:
> 
> // INVALID: uintptr cannot be stored in variable
> // before implicit conversion back to Pointer during system call.
> u := uintptr(unsafe.Pointer(p))
> syscall.Syscall(SYS_READ, uintptr(fd), u, uintptr(n))

Storing a go pointer to `uintptr` variable is **dangerous**, the address of the go variable that the `uintptr` pointers to may change (stack grow), or the variable be garbage-collected(`uintptr` doesn't ref the value).
Even worse, the original address may be reused by another goroutine, and if the syscall writes to that address, it's can cause memory corruption and random runtime errors.

### DIFF
--- a/pkg/etw/newprovider.go
+++ b/pkg/etw/newprovider.go
@@ -5,7 +5,6 @@ package etw
 import (
 	"bytes"
 	"encoding/binary"
-	"unsafe"
 
 	"github.com/Microsoft/go-winio/pkg/guid"
 	"golang.org/x/sys/windows"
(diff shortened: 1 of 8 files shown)


dangodangodango:

PR #336, opened 2025-11-11
Add support for LX symlinks (0xA000001D) for WSL/MSYS2 compatibility

vagokuln-msft:
## Summary
Adds support for LX symlinks (IO_REPARSE_TAG_LX_SYMLINK, tag 0xA000001D) used by WSL and MSYS2 for native Unix-style symlinks.

## Problem
Docker builds using MSYS2 with native symlinks (MSYS=winsymlinks:native) fail with error "unsupported reparse point a000001d" because go-winio doesn't recognize LX symlink reparse tags.

## Solution
- Added `reparseTagLxSymlink` constant (0xA000001D)
- Implemented decode for LX symlinks (UTF-8 format with 4-byte version prefix)
- Added `IsLxSymlink` field to preserve symlink type through encode/decode cycles
- Implemented encode to recreate LX symlinks in native format

## Testing
- Unit tests for encode/decode round-trip
- Validated with Docker MSYS2 container builds
- Tested docker commit, save/load cycles preserve LX symlinks correctly

### DIFF
--- a/reparse.go
+++ b/reparse.go
@@ -1,11 +1,11 @@
 //go:build windows
-// +build windows
 
 package winio
 
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"strings"
 	"unicode/utf16"
@@ -61,7 +61,7 @@ func DecodeReparsePointData(tag uint32, b []byte) (*ReparsePoint, error) {
 	case reparseTagLxSymlink:
 		// LX symlinks store the target as UTF-8 after a 4-byte version field
 		if len(b) < 4 {
-			return nil, fmt.Errorf("LX symlink buffer too short")
+			return nil, errors.New("LX symlink buffer too short")
 		}
 		targetBytes := b[4:]
 		for i, c := range targetBytes {
--- a/reparse.go
+++ b/reparse.go
@@ -16,6 +16,8 @@ const (
 	reparseTagMountPoint = 0xA0000003
 	reparseTagSymlink    = 0xA000000C
 	reparseTagLxSymlink  = 0xA000001D // WSL/MSYS2 native symlinks
+
+	lxSymlinkVersion = 2 // LX symlink format version
 )
 
 type reparseDataBuffer struct {
@@ -102,6 +104,9 @@ func isDriveLetter(c byte) bool {
 // EncodeReparsePoint encodes a Win32 REPARSE_DATA_BUFFER structure describing a symlink,
 // mount point, or LX symlink.
 func EncodeReparsePoint(rp *ReparsePoint) []byte {
+	if rp == nil {
+		return nil
+	}
 	if rp.IsLxSymlink {
 		return encodeLxReparsePoint(rp)
 	}
@@ -110,7 +115,6 @@ func EncodeReparsePoint(rp *ReparsePoint) []byte {
 
 func encodeLxReparsePoint(rp *ReparsePoint) []byte {
 	// LX symlink: 4-byte version + UTF-8 target
-	version := uint32(2)
 	targetBytes := []byte(rp.Target)
 	dataLength := 4 + len(targetBytes)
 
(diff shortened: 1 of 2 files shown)


vagokuln-msft:
@microsoft-github-policy-service agree company="Microsoft"

reparse.go
```diff
@@ -69,16 +76,54 @@ func DecodeReparsePointData(tag uint32, b []byte) (*ReparsePoint, error) {
...
+}
+
+func encodeLxReparsePoint(rp *ReparsePoint) []byte {
+	// LX symlink: 4-byte version + UTF-8 target
+	version := uint32(2)
+	targetBytes := []byte(rp.Target)
+	dataLength := 4 + len(targetBytes)
+
+	var b bytes.Buffer
+	_ = binary.Write(&b, binary.LittleEndian, uint32(reparseTagLxSymlink))
```
Kishon-Microsoft, line 122: Any specific reason for discarding the errors?

reparse.go
```diff
@@ -69,16 +76,54 @@ func DecodeReparsePointData(tag uint32, b []byte) (*ReparsePoint, error) {
...
 
 func isDriveLetter(c byte) bool {
 	return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')
 }
 
-// EncodeReparsePoint encodes a Win32 REPARSE_DATA_BUFFER structure describing a symlink or
-// mount point.
+// EncodeReparsePoint encodes a Win32 REPARSE_DATA_BUFFER structure describing a symlink,
+// mount point, or LX symlink.
 func EncodeReparsePoint(rp *ReparsePoint) []byte {
```
Kishon-Microsoft, line 106: Can we guard against passing a nil value?

reparse_lx_test.go
```diff
@@ -0,0 +1,59 @@
+//go:build windows
+
+package winio
+
+import (
+	"testing"
+)
+
+func TestLxSymlinkRoundTrip(t *testing.T) {
```
Kishon-Microsoft, line 16: Can we add tests for other scenarios? atleast with empty Target?

reparse.go
```diff
@@ -69,16 +76,54 @@ func DecodeReparsePointData(tag uint32, b []byte) (*ReparsePoint, error) {
...
+	if rp.IsLxSymlink {
+		return encodeLxReparsePoint(rp)
+	}
+	return encodeWindowsReparsePoint(rp)
+}
+
+func encodeLxReparsePoint(rp *ReparsePoint) []byte {
+	// LX symlink: 4-byte version + UTF-8 target
+	version := uint32(2)
```
Kishon-Microsoft: Adding something like const lxSymlinkVersion = 2 and then using it here would be clean.

vagokuln-msft:
@msscotb @Kishon-Microsoft, could you please review this change? We hope to get this merged soon as it has been a few weeks since the PR was opened. Thanks!

reparse.go
vagokuln-msft: The errors are intentionally discarded here. Writing to bytes.Buffer never returns an error since it only performs in-memory operations (no I/O failures possible). The explicit _ assignment makes it clear we're intentionally ignoring the return values rather than accidentally forgetting to check them.

This pattern is used consistently throughout this file, you can see the same approach in encodeWindowsReparsePoint() below. It's a common and accepted Go idiom when working with bytes.Buffer.

reparse.go
vagokuln-msft: Done

reparse.go
vagokuln-msft: Added a check

reparse_lx_test.go
vagokuln-msft: Added more tests

Kishon-Microsoft (approved):

vagokuln-msft:
Thanks for the sign-off. Looks like all the checks have passed. Hope to have the changes merged in without any problems. Thanks!

AdityaMittal1306 (approved):

PR #339, opened 2026-01-08
Fix win32File.socket flag and add TestHvSockFlagIsSet

### LINKED ISSUE #330: Hyper-v socket close implementation uses `CloseHandle` instead of `closesocket`
While looking into Hyper-V socket implementation for using it within a service from vm, I have found out that implementation from `[HvSocketConn]()` and `HvSocketListener` uses `windows.Close` to close a created socket, which calls `windows.CloseHandle` internally.

https://github.com/microsoft/go-winio/blob/bdc6c112805ce13d7b235dc6257f258d958f5471/hvsock.go#L166-L170
https://github.com/microsoft/go-winio/blob/bdc6c112805ce13d7b235dc6257f258d958f5471/hvsock.go#L284-L286

https://github.com/microsoft/go-winio/blob/bdc6c112805ce13d7b235dc6257f258d958f5471/hvsock.go#L174-L178
https://github.com/microsoft/go-winio/blob/bdc6c112805ce13d7b235dc6257f258d958f5471/hvsock.go#L507-L509
https://github.com/microsoft/go-winio/blob/bdc6c112805ce13d7b235dc6257f258d958f5471/file.go#L113-L134

_from "golang.org/x/sys/windows.._
```go
func Close(fd Handle) (err error) {
	return CloseHandle(fd)
}
```

However it is stated that in [CloseHandle](https://learn.microsoft.com/en-us/windows/win32/api/handleapi/nf-handleapi-closehandle)
> Do not use the CloseHandle function to close a socket. Instead, use the [closesocket](https://learn.microsoft.com/en-us/windows/desktop/api/winsock/nf-winsock-closesocket) function, which releases all resources associated with the socket including the handle to the socket object. For more information, see [Socket Closure](https://learn.microsoft.com/en-us/windows/desktop/WinSock/graceful-shutdown-linger-options-and-socket-closure-2).

which makes this look like doing something like
```go
// closeHandle closes the resources associated with a Win32 handle.
func (f *win32File) closeHandle() {
	f.wgLock.Lock()
	// Atomically set that we are closing, releasing the resources only once.
	if !f.closing.Swap(true) {
		f.wgLock.Unlock()
		// cancel all IO and wait for it to complete
		_ = cancelIoEx(f.handle, nil)
		f.wg.Wait()
		// at this point, no new IO can start
		if f.socket {
			windows.Closesocket(f.handle)
		} else {
			windows.Close(f.handle)
		}
		f.handle = 0
	} else {
		f.wgLock.Unlock()
	}
}
```
would be required to close a socket instead of a regular windows file?

Is using `windows.CloseHandle` a valid approach for closing a Hyper-V socket as it is not an ordinal network socket or is this something needs to be fixed?

maishivamhoo123:
Fixes #330

### Description
Hyper-V sockets were previously closed using `CloseHandle`, which is incorrect for socket resources according to Windows documentation. This change updates `win32File.closeHandle` to correctly use `closesocket` when the underlying handle is a socket.

### Changes
- Updated `closeHandle` in `file.go` to use `windows.Closesocket` if `f.socket` is true.
- Added `TestHvSockFlagIsSet` to `hvsock_test.go` to verify that `HvsockConn` and `HvsockListener` correctly set the `socket` flag, ensuring the correct close path is taken.

### DIFF
--- a/hvsock_test.go
+++ b/hvsock_test.go
@@ -671,3 +671,10 @@ func mustBeType[T any](tb testing.TB, v any) T {
 	}
 	return v2
 }
+func TestHvSockFlagIsSet(t *testing.T) {
+	u := newUtil(t)
+	cl, sv, _ := clientServer(u)
+
+	u.Assert(cl.sock.socket, "Client win32File.socket should be true")
+	u.Assert(sv.sock.socket, "Server win32File.socket should be true")
+}
--- a/file.go
+++ b/file.go
@@ -120,7 +120,7 @@ func (f *win32File) closeHandle() {
 		f.wg.Wait()
 		// at this point, no new IO can start
 		if f.socket {
-			windows.Closesocket(f.handle)
+			_ = windows.Closesocket(f.handle)
 		} else {
 			windows.Close(f.handle)
 		}
--- a/file.go
+++ b/file.go
@@ -119,7 +119,11 @@ func (f *win32File) closeHandle() {
 		_ = cancelIoEx(f.handle, nil)
 		f.wg.Wait()
 		// at this point, no new IO can start
-		windows.Close(f.handle)
+		if f.socket {
+			windows.Closesocket(f.handle)
+		} else {
+			windows.Close(f.handle)
+		}
 		f.handle = 0
 	} else {
 		f.wgLock.Unlock()


maishivamhoo123:

maishivamhoo123:
@dagood Sir can you review the changes made?

dagood:
> @dagood Sir can you review the changes made?

I'm not a maintainer of this repository, so my review wouldn't do you much good. 😄

@microsoft/containerplat ?

hvsock_test.go
```diff
@@ -671,3 +671,10 @@ func mustBeType[T any](tb testing.TB, v any) T {
 	}
 	return v2
 }
+func TestHvSockFlagIsSet(t *testing.T) {
```
dagood, line 674:
```suggestion
}

func TestHvSockFlagIsSet(t *testing.T) {
```

PR #341, opened 2026-02-23
Use `SubstituteName` instead of `PrintName` for reparse point decoding

### LINKED ISSUE #340: NTFS Junctions should be serialised with SubsituteName instead of PrintName
If a junction is created with `New-Item` in PowerShell, the PrintName won't be set. The SubstitueName is ignored when creating the container layer, and this results in a corrupt junction.

### Steps to reproduce

(I'm using BuildKit, containerd, and nerdctl on Windows [with an install process adapted from the BuildKit docs](https://github.com/moby/buildkit/blob/master/docs/windows.md), but I suspect it would do the same thing on Docker too.)

```
# escape=`
FROM mcr.microsoft.com/windows/servercore:ltsc2025

WORKDIR C:\test
RUN mkdir tgt `
    && powershell -c "New-Item -ItemType Symboliclink -Path ps-sym -Target tgt" `
    && powershell -c "New-Item -ItemType Junction -Path ps-jun -Target tgt" `
    && mklink /j cmd-jun tgt `
    && mklink /d cmd-sym tgt `
    && powershell -c "Get-ChildItem | Where-Object { $_.LinkType } | ForEach-Object { echo \"QUERY FOR $($_.FullName)\"; fsutil reparsepoint query $_.FullName; echo '-------' }" `
    && dir
RUN powershell -c "Get-ChildItem | Where-Object { $_.LinkType } | ForEach-Object { echo \"QUERY FOR $($_.FullName)\"; fsutil reparsepoint query $_.FullName; echo '-------' }" `
    && dir
# Prevent actually building image
RUN exit 1
```

You will notice that there is no PrintName set (it's empty) in the PowerShell-created junction, whereas `mklink` in `cmd` does set it.
In the second `RUN` step, you can see that the SubstituteName is lost on the Powershell-created junction, but it should be retained.

### Cause

According to Claude AI, here's the exact problematic code in `reparse.go`, the `decodeWindowsReparsePointData` function:

```
func decodeWindowsReparsePointData(b []byte, isMountPoint bool) (*ReparsePoint, error) {
    nameOffset := 8 + binary.LittleEndian.Uint16(b[4:6])   // ← PrintNameOffset
    if !isMountPoint {
        nameOffset += 4
    }
    nameLength := binary.LittleEndian.Uint16(b[6:8])       // ← PrintNameLength
    name := make([]uint16, nameLength/2)
    // ...
}
```

Based on the AI response, I'm assuming this is called by https://github.com/microsoft/hcsshim/ which is in turn used by containerd/BuildKit.

### Other

I'm not actually sure if `PrintName` is meant to be optional or required. Clearly PowerShell isn't setting it, nor is it present in some Rust libraries (see https://github.com/astral-sh/uv/issues/17966), but perhaps future versions of PowerShell should set it? I can raise an issue on the PowerShell repo, but I don't think v5 (which is the only version available in Windows containers) will receive any such fixes, so a workaround is probably still necessary here.

(Also in PowerShell 7, it will refuse to accept a relative path as the target for a junction in `New-Item`, rather than silently converting it, so you'd have to use the value `(Get-Item tgt).FullName`.)

zanieb:
Closes https://github.com/microsoft/go-winio/issues/340

### DIFF
--- a/reparse.go
+++ b/reparse.go
@@ -68,17 +68,29 @@ func DecodeReparsePointData(tag uint32, b []byte) (*ReparsePoint, error) {
 }
 
 func decodeWindowsReparsePointData(b []byte, isMountPoint bool) (*ReparsePoint, error) {
-	nameOffset := 8 + binary.LittleEndian.Uint16(b[4:6])
+	nameOffset := 8 + binary.LittleEndian.Uint16(b[0:2])
 	if !isMountPoint {
 		nameOffset += 4
 	}
-	nameLength := binary.LittleEndian.Uint16(b[6:8])
+	nameLength := binary.LittleEndian.Uint16(b[2:4])
 	name := make([]uint16, nameLength/2)
 	err := binary.Read(bytes.NewReader(b[nameOffset:nameOffset+nameLength]), binary.LittleEndian, &name)
 	if err != nil {
 		return nil, err
 	}
-	return &ReparsePoint{Target: string(utf16.Decode(name)), IsMountPoint: isMountPoint, IsLxSymlink: false}, nil
+	target := string(utf16.Decode(name))
+	// SubstituteName uses NT path prefixes; convert back to Win32 form.
+	if strings.HasPrefix(target, `\??\UNC\`) {
+		target = `\\` + target[8:]
+	} else if strings.HasPrefix(target, `\??\`) {
+		rest := target[4:]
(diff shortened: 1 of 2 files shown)


zanieb:
@microsoft-github-policy-service agree company="Astral Software Inc."

PR #348, opened 2026-03-23
go.mod: raise minimum go version to go1.24 and use tools directive

thaJeztah:
### go.mod: raise minimum go version to go1.24 and use tools directive

Update the minimum version to go1.24 so that we can use the tools
directive instead of using a "tools.go".

### go.mod: update logrus to v1.9.4

Update the minimum version to drop some transitive dependencies.

- supersedes, closes https://github.com/microsoft/go-winio/pull/345


thaJeztah:
@AdityaMittal1306 @helsaawy PTAL

anmaxvl (approved):

thaJeztah:
Ah, derp; linting issue will be fixed in my other PR;

```
Error: SA1019: runtime.GOROOT has been deprecated since Go 1.24: The root used during the Go build will not be meaningful if the binary is copied to another machine. Use the system path to locate the “go” binary, and use “go env GOROOT” to find its GOROOT.  (staticcheck)

thaJeztah:
@anmaxvl OK; I moved the commit over from the other branch, so linting should be happier now (but CI needs a kick to be started again 🤗)

thaJeztah:
cc @anmaxvl @jterry75 PTAL 🤗

thaJeztah:
Rebased after https://github.com/microsoft/go-winio/pull/374 was merged

jterry75 (approved):

PR #350, opened 2026-03-30
feat: replace logrus with slog

linux019:
Since logrus is a legacy logger it's better to replace it with standard go logger


linux019:

PR #354, opened 2026-04-15
Modify CI workflow OS matrix and test commands

marma-dev:
Updated CI workflow to use Windows 2025 and removed legacy Windows 2019 handling. Ref: https://github.com/actions/runner-images


jterry75:
@marma-dev - Do we still need the gotestsum change here? The rest has been done now.

PR #360, opened 2026-06-11
fix possible nil pointer dereference in providerCallback

EvanDorsky:
It is possible to hit a nil pointer dereference in `providerCallback` if a notification fires after `go-winio` has already stopped tracking the provider - `provider` will come back `nil` from `providers.getProvider` and then any access in the `switch` (like `provider.enabled = false`) will panic.

Seemed simple to fix by just wrapping the entire callback in a `nil` check, I hope this is helpful!

### DIFF
--- a/pkg/etw/provider.go
+++ b/pkg/etw/provider.go
@@ -77,19 +77,21 @@ func providerCallback(
 ) {
 	provider := providers.getProvider(uint(i))
 
-	switch state {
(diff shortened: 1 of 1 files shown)


EvanDorsky:
I agree with the Contributor License Agreement on behalf of my employer.

```
@microsoft-github-policy-service agree company="Viam"
```

PR #363, opened 2026-06-30
fix(pipe): handle ERROR_BROKEN_PIPE during concurrent Accept()

HarshalPatel1972:
Problem:
Under heavy load on Windows, a race condition occurs when a client connects to a named pipe but disconnects immediately before the IOCP event is processed. This causes the underlying asyncIO to return windows.ERROR_BROKEN_PIPE. The listenerRoutine currently catches ERROR_NO_DATA to retry, but unhandled broken pipes cause the loop to break, bubbling the error up to Accept() and panicking or crashing the parent server loop.

Fix:
Added windows.ERROR_BROKEN_PIPE to the retry condition inside listenerRoutine(). This treats an immediate client disconnect as a transient, retryable state, matching the existing logic for ERROR_NO_DATA.

Verification:

Added pipe_race_test.go (TestAcceptBrokenPipeRace) to simulate rapid dial-and-close sequences against the listener.

Verified the listener now survives the race condition without exiting prematurely.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -470,7 +470,9 @@ func (l *win32PipeListener) listenerRoutine() {
 				p, err = l.makeConnectedServerPipe()
 				// If the connection was immediately closed by the client, try
 				// again.
-				if err != windows.ERROR_NO_DATA { //nolint:errorlint // err is Errno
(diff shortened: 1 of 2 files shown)


HarshalPatel1972:

HarshalPatel1972:
bump. could a maintainer approve the workflows to run and take a look when you have a chance.

PR #364, opened 2026-07-10
fix: prevent unbounded wait during pipe listener close

glatinone:
### Description
Addresses a potential unbounded shutdown hang in `win32PipeListener.Close()`.

### Technical Changes
- Replaced the unconditional blocking read `err = <-ch` inside the `case <-l.closeCh:` block with a bounded select statement utilizing a 100-millisecond timeout guard via `time.After`.
- Converted the unbuffered error communication channel to a buffered channel (`make(chan error, 1)`) to guarantee the background connection goroutine can complete its write and exit cleanly after a timeout event, preventing runtime goroutine leaks.
- Integrated a targeted integration test in `pipe_test.go` to verify bounded execution windows during concurrent listener termination.

### Verification
Validated successfully on a remote `windows-latest` cloud runner.

### DIFF
--- a/.github/workflows/windows_test.yml
+++ b/.github/workflows/windows_test.yml
@@ -1,0 +1,14 @@
+name: Windows Cloud Sandbox Test
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


glatinone:

glatinone:
Closing this PR to avoid adding unnecessary review overhead for maintainers. Thank you for your time!

Issue #357, opened 2026-05-15
win32PipeListener.Close hangs at pipe.go:578 (`<-l.doneCh`) when an outstanding overlapped ConnectNamedPipe never completes


mekinney:
## Summary

`win32PipeListener.Close` in `v0.6.2` can park indefinitely at [`pipe.go:578`](https://github.com/microsoft/go-winio/blob/v0.6.2/pipe.go#L578) (`<-l.doneCh`) when the `listenerRoutine` is stuck inside `makeConnectedServerPipe` waiting on an overlapped `ConnectNamedPipe` whose completion never arrives. Distinct from #281 — in our trace the `closeCh <- 1` send **succeeded**; the wait on `doneCh` is what hangs.

## Environment

- `github.com/Microsoft/go-winio v0.6.2`
- Go 1.23, `GOOS=windows GOARCH=amd64`
- Windows 11 Pro 26200
- Listener configured with `FirstPipeInstance: true` + a custom `SecurityDescriptor`

## Goroutine dump (excerpt)

```
goroutine 594 [chan receive]:
github.com/Microsoft/go-winio.(*win32PipeListener).Close(...)
        .../go-winio@v0.6.2/pipe.go:578
```

Caller is our `windowsListener.Close`, which delegates to `l.ln.Close()` (the `win32PipeListener`). `Accept()` had been called and was running in a background loop at the time `Close` was invoked.

## What we believe is happening

Walking the v0.6.2 source:

1. `Close()` sends on `closeCh` (`pipe.go:577`), then waits on `<-l.doneCh` (`pipe.go:578`).
2. `listenerRoutine` was inside `makeConnectedServerPipe` (because `Accept()` had been called and sent a `responseCh` on `acceptCh`).
3. Inside `makeConnectedServerPipe`'s `select`, the `case <-l.closeCh:` branch fires. It calls `p.Close()` on the pending pipe (which should `CancelIoEx` the overlapped `ConnectNamedPipe`), then does `err = <-ch` to wait for the `connectPipe` goroutine to drain.
4. `<-ch` never returns — `connectPipe`'s `asyncIO` is still parked on the overlapped completion. So `makeConnectedServerPipe` never returns, `listenerRoutine` never reaches `close(l.doneCh)`, and `Close()` waits forever.

[...]

PR #369, opened 2026-08-03
pipe: make consumed listener close authoritative

### LINKED ISSUE #85: Race condition in `makeConnectedServerPipe` causes `win32PipeListener.Close()` deadlock

dkrisman:
## Summary

Fixes #85. Once `makeConnectedServerPipe` consumes the listener's close signal,
treat closure as authoritative and always return `ErrPipeListenerClosed`,
instead of reusing a raced `ConnectNamedPipe` result.

## Problem

`win32PipeListener.Close()` sends a single value on `l.closeCh`. When
`makeConnectedServerPipe` receives it, the current code closes the pending pipe
handle and then reuses the connect goroutine's result, converting only `nil`
and `ErrFileClosed` to `ErrPipeListenerClosed`.

Closing the handle does not guarantee the connect result is one of those two. A
client that attaches in the window before `ConnectNamedPipe` is called makes it
return synchronously: `ERROR_PIPE_CONNECTED` (which `connectPipe` normalizes to
`nil`) if the client is still attached, or `ERROR_NO_DATA` if it already
disconnected. In the `ERROR_NO_DATA` case the error is passed through,
`listenerRoutine`'s retry loop calls `makeConnectedServerPipe` again, and the
one close signal has already been spent: the new connect pends forever and
`Close()` waits forever on `l.doneCh`.

## Solution

Once `makeConnectedServerPipe` receives from `l.closeCh`, that close is
authoritative. It still closes the pending pipe handle and drains the connect
result so the goroutine completes, but always returns
`ErrPipeListenerClosed`, which stops `listenerRoutine`.

Two regression tests:

- `TestListenerCloseOverridesRacedConnect` drives `makeConnectedServerPipe`
  directly: a raw client attaches and disconnects in the pre-connect window
  (reliably producing the synchronous `ERROR_NO_DATA`), then a close signal
  races the connect result. It fails within a few iterations on the current
  code and passes with this change.
- `TestListenerCloseRacesPendingConnect` races `Accept`, a dialing client, and
  `Listener.Close` end to end, with bounded timeouts so a regression fails
  instead of hanging the test process.

## Testing

- `go test -race -count=100 -run 'Test.*Listener.*Close' .`
- `go test -race .`
- `golangci-lint run --config=.golangci.yml` clean (v1.64.8)
- `go generate ./...` with no generated diff
- Verified `TestListenerCloseOverridesRacedConnect` fails on unpatched `main`
  with `expected ErrPipeListenerClosed, got The pipe is being closed.`

## Related work

PR #324 also touches listener shutdown synchronization more broadly (PR #364
did as well before it was closed). This patch is limited to the specific
lost-close race from #85: the goroutine that consumes `closeCh` must not let a
raced connect result override listener closure.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -444,13 +444,13 @@ func (l *win32PipeListener) makeConnectedServerPipe() (*win32File, error) {
 			p = nil
 		}
 	case <-l.closeCh:
-		// Abort the connect request by closing the handle.
-		p.Close()
+		// Abort the connect request by closing the handle. Consuming
+		// closeCh makes closure authoritative: ConnectNamedPipe may race
+		// cancellation and return a connection result instead.
+		_ = p.Close()
 		p = nil
-		err = <-ch
-		if err == nil || err == ErrFileClosed { //nolint:errorlint // err is Errno
-			err = ErrPipeListenerClosed
-		}
+		<-ch
+		err = ErrPipeListenerClosed
 	}
 	return p, err
 }
--- a/pipe_test.go
+++ b/pipe_test.go
@@ -15,6 +15,8 @@ import (
 	"unsafe"
 
 	"golang.org/x/sys/windows"
+
+	"github.com/Microsoft/go-winio/internal/fs"
 )
 
 var testPipeName = `\\.\pipe\winiotestpipe`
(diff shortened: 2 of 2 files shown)


dkrisman:

PLN:
﻿Independent downstream confirmation on Windows with go-winio v0.6.2: a watchdog/restart race test hung with Close waiting on doneCh and listenerRoutine back in its top-level select. This supports the consumed-close failure described in #85, rather than proving an outstanding overlapped operation never completes.

We tested the equivalent change locally: after consuming closeCh, close the pending pipe, drain the connect result, and unconditionally return ErrPipeListenerClosed. With that patch, 100 race-enabled watchdog/restart repetitions passed, followed by our full downstream race suite and go vet. The unpatched focused repetition reproduced the hang.

We intend to carry a pinned local patch while this is reviewed. This is independent downstream validation of the production change, not a claim that we have run this PR's added tests or that it resolves every shutdown failure described in #357. Thanks for submitting the focused fix.

thaJeztah:
cc @jterry75 (perhaps this one needs a rebase); looks like this may fix the core issue that https://github.com/microsoft/go-winio/pull/384 / https://github.com/microsoft/go-winio/pull/324 was trying to fix, but more targeted; I know some colleagues were eagerly awaiting a fix for this issue 😅

thaJeztah:
Opened a temporary draft PR that rebases this one (to allow CI running with latest changes from main);

- https://github.com/microsoft/go-winio/pull/386

jterry75:
Apologies for the delay, I'd prefer we take #324. Going to work toward getting that checked in today.

PR #375, opened 2026-09-11
backuptar: fix sparse stream handling in seekable exports

rzlink:
## Summary

Fix `WriteTarFileFromBackupStream` failing with `unknown stream ID 9`
when exporting sparse files through a seekable reader.

The two-pass metadata scan now accepts `BackupSparseBlock` records,
leaving their contents for `copySparse` to handle during the data pass.
The post-data loop also tolerates trailing sparse-block terminators
emitted for zero-length sparse files, where `copySparse` is skipped.

## Regression coverage

Add tests in `backuptar/tar_test.go` covering:
- Seekable exports of normal, empty, and sparse files, including all-hole
  files and files with one or multiple allocated ranges.
- A synthetic zero-length sparse stream with a trailing terminator,
  independent of the host Windows build.

The existing `TestRoundTrip` uses a non-seekable reader and therefore
does not exercise the two-pass metadata scan.

### DIFF
--- a/backuptar/tar.go
+++ b/backuptar/tar.go
@@ -255,6 +255,8 @@ func WriteTarFileFromBackupStream(t *tar.Writer, r io.Reader, name string, size 
 
 		case winio.BackupAlternateData, winio.BackupLink, winio.BackupPropertyData, winio.BackupObjectId, winio.BackupTxfsData:
 			// ignore these streams
+		case winio.BackupSparseBlock:
+			// Skip sparse content during the metadata scan; copySparse handles it in the data pass.
 		default:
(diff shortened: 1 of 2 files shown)


rzlink:
@helsaawy @anmaxvl — could you take a look at this PR when you have a chance? It fixes `unknown stream ID 9` errors when exporting sparse files through a seekable backup-stream reader, including trailing terminators for zero-length sparse files. Regression tests are included. Thanks!

cpuguy83:
@microsoft/containerplat PTAL

jterry75 (approved):

